PR #65, opened 2023-12-12
Implement routing configuration for Backstage server endpoint

jianrongzhang89:
Changes made:
1. Autodetect if the current cluster is OpenShift and creates a route when Backstage CR is reconciled.
2. Changed backstage deployment and service name to "backstage-"+ to allow the users to deploy multiple backstage CRs in the same namespace.
3. Changed the default value of own-runtien to true.
Note [#1](https://github.com/janus-idp/operator/issues/1): supporting host and tls in the route via Backstage CR is not in the scope of this PR. Such work shall be done later as a separate PR after the data model has been finalized.
Note [#2](https://github.com/janus-idp/operator/pull/2): unit test with isOpenShift=true is not included as OpenShift specific objects (routes, projects etc) are not currently supported by the envTest tool (see [operator-framework/operator-sdk#4434](https://github.com/operator-framework/operator-sdk/issues/4434)).

## Which issue(s) does this PR fix or relate to

- Fixes #53 
## PR acceptance criteria

- [ X] Tests
- [ ] Documentation

## How to test changes / Special notes to the reviewer
1) Deploy the operator  on OpenShift and create the postgres secret (make sure POSTGRES_HOST is set to "backstage-psql-<your CR name>", and create backstage CR.
2) Check that a route has been created:
oc get route
NAME            HOST/PORT                                                        PATH   SERVICES        PORT           TERMINATION     WILDCARD
backstage-bs1   backstage-bs1-backstage.apps.cluster-s4l2f.dynamic.opentlc.com   /      backstage-bs1   http-backend   edge/Redirect   None
3) Now check if you can open the page in your browser:
https://<route host>

Issue #1, opened 2023-10-31
[Epic] Operator based on the Operator SDK


jasperchui:
## Goal

This epic aims to provide an operator based on Operator SDK that is releaseable and usable state where it offers a comparable stable experience as installing via helm chart.

## Acceptance criteria

- [ ] Operator is able to deploy Backstage
- [ ] Operator can reconcile changes
- [ ] Operator can enforce Janus IDP image
- [ ] Operator can enforce proper routing

## Requirements

- [ ] Test plan
- [ ] Documentation

## Roadmap
https://github.com/orgs/janus-idp/projects/5/views/1

## Notes

**Additional context**
Add any other context or screenshots about the epic here.

rm3l:
/close

Closing as done.

Issue #63, opened 2023-12-11
Do not make creating a secret for PostgreSQL a prerequisite


rm3l:
## User Story

As an Operator user, I want to have a Backstage instance running without having to explicitly create a secret for PostgreSQL, so that I can get started as quickly as possible.
Currently, I need to create a secret named `postgres-secrets` first, which is then injected as environment variables into both the PostgreSQL StatefulSet and the Backstage Deployment. Otherwise, the pods won't be able to start.
The operator could easily generate such a secret for me.

## Acceptance Criteria

- [ ] It should not require a `postgres-secrets` Secret to be created as a prerequisite
- [ ] If `enableLocalDb` is `true` or unset (i.e., a local Database should be created), the operator should dynamically create such secret (unless it exists already) with the necessary values to use in the DB and Backstage resources
- [ ] It should create random values for password fields in the Secret
- [ ] It should set the name of this secret into the CR status
- [ ] It should inject the right information (env vars) into the DB and Backstage containers
- [ ] It should not create a Postgres Secret if `enableLocalDb` is `false` (the user probably wants to use their own external DB)

gazarenkov:
A couple questions to this DoD (in addition to https://github.com/janus-idp/operator/pull/68#pullrequestreview-1779831368 ) :
> It should create random values for password fields in the Secret 

Is it really important to generate a random value for the local, namespace scoped database or static default value (which can be changed) would be sufficient? 

> It should not create a Postgres Secret if enableLocalDb is true (the user probably wants to use their own external DB)

you meant "is false"?

gazarenkov:
> > A couple questions to this DoD (in addition to [#68 (review)](https://github.com/janus-idp/operator/pull/68#pullrequestreview-1779831368) ) :
> > > It should create random values for password fields in the Secret
> > 
> > 
> > Is it really important to generate a random value for the local, namespace scoped database or static default value (which can be changed) would be sufficient?
> 
> I think it might be important to have different creds for each DB (at least from a security perspective). I guess different people could work on a shared namespace and have multiple Backstage instances in that namespace, right? 

Well, it is a good question. TBH I do not see a lot of sense in making several instances in one namespace (we do not prevent it though). And, anyway, even if several instances, do you expect them to be production ready or we are talking about development case (dummy data anyway)? (I tend to think it is second case?)

[...]

PR #67, opened 2023-12-12
Implement routing configuration for Backstage server endpoint

jianrongzhang89:
Changes made:

Autodetect if the current cluster is OpenShift and creates a route when Backstage CR is reconciled.
Changed backstage deployment and service name to "backstage-"+ to allow the users to deploy multiple backstage CRs in the same namespace.
Changed the default value of own-runtien to true.
Note https://github.com/janus-idp/operator/issues/1: supporting host and tls in the route via Backstage CR is not in the scope of this PR. Such work shall be done later as a separate PR after the data model has been finalized.
Note https://github.com/janus-idp/operator/pull/2: unit test with isOpenShift=true is not included as OpenShift specific objects (routes, projects etc) are not currently supported by the envTest tool (see https://github.com/operator-framework/operator-sdk/issues/4434).

- Fixes #65 

## PR acceptance criteria

- [ X] Tests
- [ ] Documentation

Deploy the operator on OpenShift and create the postgres secret (make sure POSTGRES_HOST is set to "backstage-psql-", and create backstage CR.
Check that a route has been created:
oc get route
NAME HOST/PORT PATH SERVICES PORT TERMINATION WILDCARD
backstage-bs1 backstage-bs1-backstage.apps.cluster-s4l2f.dynamic.opentlc.com / backstage-bs1 http-backend edge/Redirect None
Now check if you can open the page in your browser using the route host, for instance,
https://backstage-bs1 backstage-bs1-backstage.apps.cluster-s4l2f.dynamic.opentlc.com

### DIFF
--- a/examples/postgres-secret.yaml
+++ b/examples/postgres-secret.yaml
@@ -9,4 +9,4 @@ stringData:
   POSTGRES_PORT: "5432"
   POSTGRES_USER: postgres
   POSTGRESQL_ADMIN_PASSWORD: admin123
-  POSTGRES_HOST: backstage-psql
+  POSTGRES_HOST: backstage-psql-bs1
--- a/bundle/manifests/backstage-default-config_v1_configmap.yaml
+++ b/bundle/manifests/backstage-default-config_v1_configmap.yaml
@@ -142,7 +142,7 @@ data:
     apiVersion: apps/v1
     kind: Deployment
     metadata:
-      name: backstage
+      name:  # placeholder for 'backstage-<cr-name>'
     spec:
       replicas: 1
       selector:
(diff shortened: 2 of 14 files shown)


examples/postgres-secret.yaml
```diff
@@ -9,4 +9,4 @@ stringData:
   POSTGRES_PORT: "5432"
   POSTGRES_USER: postgres
   POSTGRESQL_ADMIN_PASSWORD: admin123
-  POSTGRES_HOST: backstage-psql
\ No newline at end of file
+  POSTGRES_HOST: backstage-psql-bs1
```
rm3l, line 12: In the previous PR (https://github.com/janus-idp/operator/pull/53/files#r1422807856), you were dynamically injecting a `POSTGRES_HOST` env var with the actual DB service name.
I guess you plan to work on that as part of https://github.com/janus-idp/operator/issues/63, right?

jianrongzhang89: Yes this will be done in #63 instead.

rm3l (approved):

Issue #21, opened 2023-11-16
Define (initial) set of spec.parameters for overriding default/raw runtime configuration


gazarenkov:
We initially have default (easy) and row (complex) runtime configuration options.
We clearly needed "something in between" so user is able to specify some "important" parameters.
The goal is to specify this parameters.

PR #68, opened 2023-12-13
Do not make creating a secret for PostgreSQL a prerequisite

### LINKED ISSUE #63: Do not make creating a secret for PostgreSQL a prerequisite

jianrongzhang89:
## Description
Changes:
1) Check the backstage deployment and statefulset configuration, if the main container has an envFrom entry
pointing to the secret reference with special name '{POSTGRESQL_SECRET}', a secret is auto generated, and the envFrom entry is updated to backstage-psql-secret-<CR name>. Otherwise, we assume that the user has an existing secret provided and it will not be auto generated.
2) Add a new field existingDbSecret to the backstage CRD to explicitly indicate an existing database secret is used (this is the same way as the helm chart does).

- Fixes #63 

## PR acceptance criteria

- [X ] Tests
- [ ] Documentation

a) Deploy the operator and the dummy backstage CR (examples/bs1.yaml).
b) Check that the psql db and backstage are running.
c) Check that a secret backstage-psql-secret-<CR name> has been created.
d) Delete the backstage CR and check and confirm that the secret gets deleted.
e)deploy examples/bs-existing-secret.yaml and check the backstage and plsql db are using the existing secret used without generating a new one.


rm3l (changes requested):
> The generated secret does not have own-runtime set, ie, when the backstage CR is deleted, the secret will NOT be removed. This prevents the secret from being re-created should the CR is deleted and then redeployed.

About this, I'm wondering if we should not delete the generated secret as well. Otherwise, we might end up with a lot of secrets, especially if I delete my CR, then rename it before applying it. The data in the secret is random, so I guess it should not be a problem to recreate it, no?

gazarenkov (changes requested):
A couple notes:

I think, for the sake of consistency we need to take DB secret as a first class citizen object i.e:
- it has to have own (db-secret.yaml) default configuration and accordingly to be able to be overlayed with raw configuration
- (optionally if we decide so) it has to have dedicated Backstage CRD config (just secret name)

I do not think there are a lot of reasons to generate it, static initialization should be fine IMO (note it is a property of "local", namespace scoped Database, production is a different story)

jianrongzhang89:
> > The generated secret does not have own-runtime set, ie, when the backstage CR is deleted, the secret will NOT be removed. This prevents the secret from being re-created should the CR is deleted and then redeployed.
> 
> About this, I'm wondering if we should not delete the generated secret as well. Otherwise, we might end up with a lot of secrets, especially if I delete my CR, then rename it before applying it. The data in the secret is random, so I guess it should not be a problem to recreate it, no?

Fair enough. Accepted.

jianrongzhang89:
> A couple notes:
> 
> I think, for the sake of consistency we need to take DB secret as a first class citizen object i.e:
> 
> * it has to have own (db-secret.yaml) default configuration and accordingly to be able to be overlayed with raw configuration
> * (optionally if we decide so) it has to have dedicated Backstage CRD config (just secret name)
> 
> I do not think there are a lot of reasons to generate it, static initialization should be fine IMO (note it is a property of "local", namespace scoped Database, production is a different story)

Agreed to use the same approach for the psql secret as other objects. Meanwhile, I think it is worthwhile to generate a random password instead of a default one.

config/manager/default-config/deployment.yaml
```diff
@@ -83,7 +83,7 @@ spec:
               value: "7007"
           envFrom:
             - secretRef:
-                name: postgres-secrets
+                name: "{POSTGRESQL_SECRET}"  # will be replaced with 'backstage-psql-secrets-<cr-name>'
```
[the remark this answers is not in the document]
jianrongzhang89: Well, this is a problem we need to discuss. extraEnvs adds an secret to access the db, while the default one already has a secret ref for the db access and hence we have a conflict.

gazarenkov:
@jianrongzhang89 well, I am not against generating per se, it just looks like it may cost too much.
Other methods of Backstage installation can live w/o it so let's start simple .

config/manager/default-config/deployment.yaml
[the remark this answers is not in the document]
rm3l: IIRC, there is a rule of precedence with conflicting env vars in the Pod Spec: last one wins, and values from `env` always take precedence.
https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#environment-variables

So I guess it should be fine, even if the user adds extraEnvs: the ones they adds from secret or configmap refs will always override the default ones if names of the env vars conflicts. And if they add env, those ones will take precedence.

rm3l:
> @jianrongzhang89 well, I am not against generating per se, it just looks like it may cost too much. Other methods of Backstage installation can live w/o it so let's start simple .

IMO, it would make sense for the DB to have a random password value for each Backstage instance (for the reasons depicted in https://github.com/janus-idp/operator/issues/63#issuecomment-1854383364). 
Also, 
Not sure I understand why it may cost too much. Could you please elaborate?
Also, I think other installation methods generate a random password by default if not set explicitly. That's what the Bitnami PostgreSQL Chart (which the Backstage Helm Chart depends on) seems to do in my understanding: https://github.com/bitnami/charts/blob/main/bitnami/postgresql/templates/secrets.yaml#L11

jianrongzhang89:
> > @jianrongzhang89 well, I am not against generating per se, it just looks like it may cost too much. Other methods of Backstage installation can live w/o it so let's start simple .
> 
> IMO, it would make sense for the DB to have a random password value for each Backstage instance (for the reasons depicted in [#63 (comment)](https://github.com/janus-idp/operator/issues/63#issuecomment-1854383364)). Not sure I understand why it may cost too much. Could you please elaborate? Also, I think other installation methods generate a random password by default if not set explicitly. That's what the Bitnami PostgreSQL Chart (which the Backstage Helm Chart depends on) seems to do in my understanding: https://github.com/bitnami/charts/blob/main/bitnami/postgresql/templates/secrets.yaml#L11

Also, if we do not generate the db password, users will be unable to use local psql db for production. Although they may use external db installations, I can image some smaller sized customers may simply want to use the local psql db instead.

bundle/manifests/backstage-default-config_v1_configmap.yaml
```diff
@@ -33,203 +33,79 @@ data:
...
-                  value: "7007"
-              envFrom:
-                - secretRef:
-                    name: postgres-secrets
-              #            - secretRef:
-              #                name: backstage-secrets
-              volumeMounts:
-                - mountPath: /opt/app-root/src/dynamic-plugins-root
-                  name: dynamic-plugins-root
+  db-statefulset.yaml: "apiVersion: apps/v1\nkind: StatefulSet\nmetadata:\n  name:
```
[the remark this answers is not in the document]
jianrongzhang89: Not sure what happened to the code generator, but it is fine now.

config/manager/default-config/deployment.yaml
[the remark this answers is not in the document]
jianrongzhang89: As we discussed, we would add a new field existingDbSecret to Backstage CR.

api/v1alpha1/backstage_types.go
```diff
@@ -32,13 +32,19 @@ type BackstageSpec struct {
...
+
+	// Reference to an existing secret for database access. Required for external database access.
+	// Optional for a local database (EnableLocalDb=true) and if absent a secret will be auto generated.
+	// The secret shall include information requried for the database access, for instance:
+	// "POSTGRES_PASSWORD": "rl4s3Fh4ng3M4"
+	// "POSTGRES_PORT": "5432"
+	// "POSTGRES_USER": "postgres"
+	// "POSTGRESQL_ADMIN_PASSWORD": "rl4s3Fh4ng3M4"
+	// "POSTGRES_HOST": "backstage-psql-bs1"  # For local database, set to "backstage-psql-<CR name>".
+	ExistingDbSecret *ObjectKeyRef `json:"existingDbSecret,omitempty"`
```
rm3l, line 47: With this additional field, shall we consider having a parent `postgresql` or `db` field, like so (original idea in https://github.com/janus-idp/operator/issues/21)? It makes it more structured for the users IMO.
```yaml
spec:
  postgresql:
    enableLocal: true
    existingAuthSecretName: ""
  application:
  # -- truncated --
```

api/v1alpha1/backstage_types.go
```diff
@@ -32,13 +32,19 @@ type BackstageSpec struct {
...
-	// containing references to the Database connection information,
-	// which might be supplied as environment variables (see the ExtraEnvs field) or extra-configuration files
-	// (see the ExtraFiles field in the Application structure).
 	// +optional
 	//+kubebuilder:default=true
 	EnableLocalDb *bool `json:"enableLocalDb,omitempty"`
+
+	// Reference to an existing secret for database access. Required for external database access.
+	// Optional for a local database (EnableLocalDb=true) and if absent a secret will be auto generated.
+	// The secret shall include information requried for the database access, for instance:
```
rm3l, line 41:
```suggestion
	// The secret shall include information required for the database access, for instance:
```

api/v1alpha1/backstage_types.go
```diff
@@ -32,13 +32,19 @@ type BackstageSpec struct {
...
 	// Control the creation of a local PostgreSQL DB. Set to false if using for example an external Database for Backstage.
-	// To use an external Database, you can provide your own app-config file (see the AppConfig field in the Application structure)
-	// containing references to the Database connection information,
-	// which might be supplied as environment variables (see the ExtraEnvs field) or extra-configuration files
-	// (see the ExtraFiles field in the Application structure).
 	// +optional
 	//+kubebuilder:default=true
 	EnableLocalDb *bool `json:"enableLocalDb,omitempty"`
+
+	// Reference to an existing secret for database access. Required for external database access.
```
rm3l, line 39: I think it should be fine even if the secret is not set if `enableLocalDb` is `false`. Connection information can still be provided as extra env vars if needed. And it is also just as fine, IMO.
If the user does not provide the necessary information, the Backstage process won't be able to connect to the DB, and the Pod will end up in CrashLoopBackOff (as it would anyway even if the user had provided an existing secret but with, say, wrong host/port for example).

api/v1alpha1/backstage_types.go
rm3l: Also, all those env vars are required only if `enableLocalDb` is `true`, as those are required by the PostgreSQL image we are using.
Otherwise, IMO, there is nothing preventing users from overriding the default database settings with an app-config like [this](https://github.com/janus-idp/backstage-showcase/blob/main/app-config.production.yaml), and providing an existing DB secret with those keys:
```yaml
backend:
  database:
    connection:
      host: "${MY_HOST}"
     port: "${MY_PORT}"
     user: "${MY_USER}"
     password: "${MY_PASSWORD}"
```

gazarenkov:
> > > @jianrongzhang89 well, I am not against generating per se, it just looks like it may cost too much. Other methods of Backstage installation can live w/o it so let's start simple .
> > 
> > 
> > IMO, it would make sense for the DB to have a random password value for each Backstage instance (for the reasons depicted in [#63 (comment)](https://github.com/janus-idp/operator/issues/63#issuecomment-1854383364)). Not sure I understand why it may cost too much. Could you please elaborate? Also, I think other installation methods generate a random password by default if not set explicitly. That's what the Bitnami PostgreSQL Chart (which the Backstage Helm Chart depends on) seems to do in my understanding: https://github.com/bitnami/charts/blob/main/bitnami/postgresql/templates/secrets.yaml#L11
> 
> Also, if we do not generate the db password, users will be unable to use local psql db for production. Although they may use external db installations, I can image some smaller sized customers may simply want to use the local psql db instead.

About using it for small group, I do not think it is directly related to the credential generation topic
It is still possible to install database with random credentials, the same as other configuration, just provide db-secret.yaml (for example) Secret, or specify this Secret in the Backstage.spec with generated credentials (as proposed), right?

@rm3l 
>  Not sure I understand why it may cost too much. Could you please elaborate?

What we propose for the case when one want to change the generated credentials? Like, ok, I go to sql server and change it (as you mentioned) but how I change the Secret if after first generation we consider it as not changeable and => ignore its changing? Or I missed something?

api/v1alpha1/backstage_types.go
gazarenkov: maybe just authSecretName?
(I think "existing" is odd)

jianrongzhang89: Sounds good. Changed to use authSecret.

api/v1alpha1/backstage_types.go
jianrongzhang89: I added the comment to the field indicating the env vars listed here are an example for Postgres Db.

jianrongzhang89:
> What we propose for the case when one want to change the generated credentials? Like, ok, I go to sql server and change it (as you mentioned) but how I change the Secret if after first generation we consider it as not changeable and => ignore its changing? Or I missed something?

@gazarenkov This is a good point. After the secret is generated, the user shall be able to  change the passwords in the server, and then update the secret accordingly. Current implementation does not prevent the user from doing this.

api/v1alpha1/backstage_types.go
```diff
@@ -31,14 +31,26 @@ type BackstageSpec struct {
...
+	// Reference to a secret for database authentication. Required for external database access.
+	// Optional for a local database (EnableLocalDb=true) and if absent a secret will be auto generated.
+	// The secret shall include information used for the database access.
+	// An example for PostgreSQL DB access:
+	// "POSTGRES_PASSWORD": "rl4s3Fh4ng3M4"
+	// "POSTGRES_PORT": "5432"
+	// "POSTGRES_USER": "postgres"
+	// "POSTGRESQL_ADMIN_PASSWORD": "rl4s3Fh4ng3M4"
+	// "POSTGRES_HOST": "backstage-psql-bs1"  # For local database, set to "backstage-psql-<CR name>".
+	AuthSecret *ObjectKeyRef `json:"authSecret,omitempty"`
```
rm3l, line 53: `ObjectKeyRef` also includes an optional `key` field. And setting it in the CR won't have any effect, making it confusing to me.
To be consistent with what we did with `dynamicPluginsConfigMapName`, can we consider renaming into a simple `authSecretName` string field instead?

api/v1alpha1/backstage_types.go
```diff
@@ -31,14 +31,26 @@ type BackstageSpec struct {
 	// Raw Runtime Objects configuration. For Advanced scenarios.
 	RawRuntimeConfig RuntimeConfig `json:"rawRuntimeConfig,omitempty"`
 
+	// Configuration for database access. Optional.
+	DatabaseConfig DatabaseConfig `json:"databaseConfig,omitempty"`
```
rm3l, line 35: I know that naming things is one of hardest things in computing ^^, but how about just `database`? To be consistent with `application` (and make it understandable what the components of a default Backstage instance are)

jianrongzhang89: Yes that's good.

api/v1alpha1/backstage_types.go
jianrongzhang89: Sounds good.

gazarenkov:
> > The generated secret does not have own-runtime set, ie, when the backstage CR is deleted, the secret will NOT be removed. This prevents the secret from being re-created should the CR is deleted and then redeployed.
> 
> About this, I'm wondering if we should not delete the generated secret as well. Otherwise, we might end up with a lot of secrets, especially if I delete my CR, then rename it before applying it. The data in the secret is random, so I guess it should not be a problem to recreate it, no?

We are talking about the case when Operator generates the secret right?
In this case the generated secret is a part of operator.
So, when own-runtime = true (normally) it has to be fully synchronized with CR, i.e. it should be deleted along with CR. 
Unlike the case when the secret is external - i.e. you point it as a part of CR.spec (including raw config), in this case operator does not own it and it will not be touched

gazarenkov:
> > What we propose for the case when one want to change the generated credentials? Like, ok, I go to sql server and change it (as you mentioned) but how I change the Secret if after first generation we consider it as not changeable and => ignore its changing? Or I missed something?
> 
> @gazarenkov This is a good point. After the secret is generated, the user shall be able to change the passwords in the server, and then update the secret accordingly. Current implementation does not prevent the user from doing this as the secret gets generated only if it does not exist yet.

In a case when secret is generated with default configuration and synchronized with operator, I do not think existence of secret is a good criteria to decide (indeed when you create CR with defaults, secret does not exist always).
I would say (and that's how it is implemented in https://github.com/gazarenkov/janus-idp-operator/tree/next) empty password data field is better criteria.

rm3l:
> > > The generated secret does not have own-runtime set, ie, when the backstage CR is deleted, the secret will NOT be removed. This prevents the secret from being re-created should the CR is deleted and then redeployed.
> > 
> > 
> > About this, I'm wondering if we should not delete the generated secret as well. Otherwise, we might end up with a lot of secrets, especially if I delete my CR, then rename it before applying it. The data in the secret is random, so I guess it should not be a problem to recreate it, no?
> 
> We are talking about the case when Operator generates the secret right? In this case the generated secret is a part of operator. So, when own-runtime = true (normally) it has to be fully synchronized with CR, i.e. it should be deleted along with CR. Unlike the case when the secret is external - i.e. you point it as a part of CR.spec (including raw config), in this case operator does not own it and it will not be touched

Yes, this was about the case when the secret was generated. I think deletion is now handled accordingly in this PR. External secrets are not deleted when the CR is deleted.

controllers/local_db_secret.go
```diff
@@ -0,0 +1,114 @@
...
+			return nil, fmt.Errorf("failed to get secret for PostgreSQL DB (%q), reason: %s", secretName, err)
+		}
+		// Create a secret with a random value
+		val := func(length int) string {
+			bytes := make([]byte, length)
+			if _, randErr := rand.Read(bytes); randErr != nil {
+				// Do not fail, but use a fallback value
+				return _defaultPostGresSecretValue
+			}
+			return string(bytes)
```
rm3l, line 65: In some cases, the generated password was not safe to use in the application, and made both the DB and Backstage pods unable to start.
After trying to decode the content in the Secret, I noticed that in some cases, we might end up with some unsafe characters, e.g.:

```
POSTGRES_PASSWORD: "\n\"��H��|W��\x03\x06uC\x05���K�i��"
```

```
$ kubectl get secrets backstage-psql-secret-bs1 -o jsonpath='{.data.POSTGRES_PASSWORD}' | base64 -d

"H|WuCKi%                                                                                                                                                                                      

```

or even:
```
POSTGRES_PASSWORD: "d옫�C\x18j��FT�w����ͦ��J�" 
```

```sh
# From the Backstage container
bash-5.1$ env | grep POSTGRES_PASS
POSTGRES_PASSWORD=d옫�Cj��FT�w����ͦ��J�
bash-5.1$ 
```

I think we'd better make it safer, maybe by further encoding it, or limiting the possible characters of the random generator to safe ones.

jianrongzhang89: Good suggestion! I added base64 encoding so that the generated password does not contain such special characters.

jianrongzhang89:
> > > What we propose for the case when one want to change the generated credentials? Like, ok, I go to sql server and change it (as you mentioned) but how I change the Secret if after first generation we consider it as not changeable and => ignore its changing? Or I missed something?
> > 
> > 
> > @gazarenkov This is a good point. After the secret is generated, the user shall be able to change the passwords in the server, and then update the secret accordingly. Current implementation does not prevent the user from doing this as the secret gets generated only if it does not exist yet.
> 
> In a case when secret is generated with default configuration and synchronized with operator, I do not think existence of secret is a good criteria to decide (indeed when you create CR with defaults, secret does not exist always). I would say (and that's how it is implemented in https://github.com/gazarenkov/janus-idp-operator/tree/next) empty password data field is better criteria.

In such case, the user should be responsible to make sure the secret is correct. Also note that the secret may not already exist when the CR is created and the user may create it after the CR is already created.

rm3l (approved):

Issue #69, opened 2023-12-13
Layered operator design


gazarenkov:
The goal of this issue is redesigning the Operator making it layered for:
- separate the internal configuration logic (the biggest part) from the K8s API calls
- better testability, implementing the absolute most testing logic as plain unit tests, decreasing amount of big and slow K8s envtests
- better documentation and, therefore supportability, to be able to express how operator is designed.

rm3l:
Priority right now is to have the first version of the operator out there for RHDH 1.1. We have something working even with all the imperfections in the codebase and tests structure, and QE/Docs will start trying it out as soon as possible.
Can we postpone this refactoring until after this first version is released?

gazarenkov:
The thing is that our initial version is mostly POC (it is not good and not bad, we just had to start somehow),  as a consequence:
- it is difficult to make Design/arch documentation since it mostly working as a plain script for creating K8s objects 
- It is difficult to support and improve for the same reason 
- it is difficult to test because logic is not layered, so all we can do is integration (envtest) tests for Controller, no unit tests which means:
     - slow run time 
     -  a lot of poorly structured code, which is difficult to understand (not being an author) and maintain
     
 So, I would still highly recommend to look at https://github.com/janus-idp/operator/pull/70 and not to postpone applying it unless there are no objections per se.
 It is not that much work to do to check and fix if needed. It already contain more tests than older one and only a couple of integration tests are not passed (commented for the time) and mostly the reason of not passing is clear.
I believe spending some time now will save a lot of time in the future.

gazarenkov:
[JIRA](https://issues.redhat.com/browse/RHIDP-1359)

PR #70, opened 2023-12-13
Refactor operator logic 

gazarenkov:
As stated in https://github.com/janus-idp/operator/issues/69 with more comments [here](https://github.com/janus-idp/operator/issues/69#issuecomment-1877032654)
the goal of this PR is to improve design, supportability and testablity of Operator.
For this purpose all the configuration complexity , forming Desired State of Custom Resource (see 3 layers of configuration) was moved out of Controller to the Model.

## Description
Configuration logic moved out from controller package to the model package, tests added accordingly.

https://github.com/janus-idp/operator/issues/69

## PR acceptance criteria

Integration and (new) unit tests passed.


config/manager/default-config/db-statefulset.yaml
```diff
@@ -26,9 +26,9 @@ spec:
               value: /var/lib/pgsql/data
             - name: PGDATA
               value: /var/lib/pgsql/data/userdata
-          envFrom:
-            - secretRef:
-                name: postgres-secrets
+#          envFrom:
+#            - secretRef:
+#                name: postgres-secrets
           image: quay.io/fedora/postgresql-15:latest
```
nickboldt, line 32: could this container image reference be defined elsewhere, like via an env var in the operator bundle's CSV?

config/manager/default-config/deployment.yaml
```diff
@@ -46,10 +53,13 @@ spec:
...
               subPath: .npmrc
+            - mountPath: /opt/app-root/src/dynamic-plugins.yaml
+              subPath: dynamic-plugins.yaml
+              name: dynamic-plugins-conf
           workingDir: /opt/app-root/src
 
       containers:
-        - name: backstage-backend
+        - name: backstage-backend    # placeholder for 'backstage-backend'
           image: quay.io/janus-idp/backstage-showcase:next
```
nickboldt, line 63: I assume this value is changed / overwritten in the downstream implementation, since we can't deploy janus-idp inside RHDH. But it seems we don't have a transformation rule for converting from upsream to downstream. Could this instead be defined as an env var in the operator bundle's CSV? then we could just have a different set of images in single place in downstream, rather than needing to grep for `image: ` in *.yaml and *.yml files ...

controllers/backstage_status.go
```diff
@@ -0,0 +1,56 @@
...
+
+package controller
+
+import (
+	"context"
+
+	bs "janus-idp.io/backstage-operator/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
```
jianrongzhang89, line 23: These methods do not need to have BackstageReconciler as the receiver, and ctx and ns are not needed for them.

api/v1alpha1/backstage_types.go
```diff
@@ -193,6 +201,66 @@ type BackstageList struct {
...
+	// caCertificate provides the cert authority certificate contents
+	CACertificate string `json:"caCertificate,omitempty"`
+}
+
 func init() {
 	SchemeBuilder.Register(&Backstage{}, &BackstageList{})
 }
+
+func (s BackstageSpec) LocalDbEnabled() bool {
+	if s.Database.EnableLocalDb == nil {
```
rm3l, line 262: How about a one-line call to `pointer.BoolDeref`?

rm3l (commented):
It is hard to test this PR thoroughly with all the conflicts reported.

image

Can you update your branch and the conflicts reported?

config/manager/default-config/deployment.yaml
gazarenkov: Why transformation and grep?
We can  have dedicated downstream, RHDP configuration (for images and for other reasons, like Backstage app configuration) and maintain it independently.

config/manager/default-config/db-statefulset.yaml
gazarenkov: It could but for the time the reason is not that clear for me.

Those (like db-statefulset.yaml) yaml files are part of configuration and this configuration is provided with bundle (see bundle/backstage-default-config_v1_configmap.yaml) along with CSV file which is located at the same bundle folder (see bundle/backstage-operator.clusterserviceversion.yaml).  
So, user (Admin) is able to change the image right here and it is not dependent on the operator version, it is a matter of configuration.
So, I am curios, why would we need to make some additional logic for envvar substitution and ask admin to maintain additional envvars instead of just modify the conf file when needed (again, it is not dependent on operator version)?

api/v1alpha1/backstage_types.go
gazarenkov: you mean: 
return pointer.BoolDeref(s.Database.EnableLocalDb, true) ?

good option as well.

controllers/backstage_status.go
gazarenkov: params fixed

api/v1alpha1/backstage_types.go
```diff
@@ -31,7 +30,7 @@ type BackstageSpec struct {
 	Application *Application `json:"application,omitempty"`
 
 	// Raw Runtime Objects configuration. For Advanced scenarios.
-	RawRuntimeConfig RuntimeConfig `json:"rawRuntimeConfig,omitempty"`
+	RawRuntimeConfig string `json:"rawRuntimeConfig,omitempty"`
```
jianrongzhang89, line 33: So BackstageConfigName and LocalDbConfigName are merged into one configmap?

config/manager/default-config/db-statefulset.yaml
```diff
@@ -27,8 +27,9 @@ spec:
             - name: PGDATA
               value: /var/lib/pgsql/data/userdata
           envFrom:
-            - secretRef:
-                name: <POSTGRESQL_SECRET>  # will be replaced with 'backstage-psql-secrets-<cr-name>'
+#            - secretRef:
```
jianrongzhang89, line 30: Why this is removed?

config/manager/default-config/deployment.yaml
```diff
@@ -81,11 +92,9 @@ spec:
...
-                name: <POSTGRESQL_SECRET>  # will be replaced with 'backstage-psql-secrets-<cr-name>'
-          #            - secretRef:
-          #                name: backstage-secrets
           volumeMounts:
             - mountPath: /opt/app-root/src/dynamic-plugins-root
-              name: dynamic-plugins-root
\ No newline at end of file
+              name: dynamic-plugins-root
+
+
```
jianrongzhang89, line 99: Remove the extra new lines

jianrongzhang89:
Due to the complexity of this refactoring PR,  this PR should be merged after other feature PRs (https://github.com/janus-idp/operator/pull/119 and may be others) are merged.

api/v1alpha1/backstage_types.go
gazarenkov: yes, it looks more consistent with default config where we have all caonfiguration in one ConfigMap

config/manager/default-config/db-statefulset.yaml
gazarenkov: this declaration is not used. Do you see any value keeping it ?

config/manager/default-config/deployment.yaml
gazarenkov: +

controllers/backstage_controller.go
```diff
@@ -99,44 +97,36 @@ func (r *BackstageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	// 1. Preliminary read and prepare some config objects from the specs (configMaps, Secrets...)
+	// 2. Make some validation to fail fast
+	spec, err := r.preprocessSpec(ctx, backstage.Spec, req.Namespace)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to preprocess backstage spec: %w", err)
 	}
 
-	err := r.applyBackstageDeployment(ctx, backstage, req.Namespace)
+	// This creates array of model objects to be reconsiled
+	bsModel, err := model.InitObjects(ctx, backstage, spec, r.OwnsRuntime, r.IsOpenShift)
```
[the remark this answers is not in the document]
gazarenkov: Not sure I understand the question
Technically statefulset and services are separate k8s objects so the most straightforward way is to check both...
Does it answer or you have in mind something else?

gazarenkov:
> Due to the complexity of this refactoring PR, this PR should be merged after other feature PRs (#119 and may be others) are merged.

I disagree,
It is a good moment to do it ASAP because:
- it synced with the current state of main
- it passed all the existing tests, so it is safe until the opposite proven
- it does not change any API and documented features
- it does not block any urgent feature requests and release plans
So, it is a complex but internal and quite long lived PR (as for our scale), which, by idea, does not impact on project functioning (unless we have other tests proving the opposite).
I do not think it is a good idea to leave it on indefinite time, this way we will need to support divergence, and so, duplicate the effort, which is not efficient I suppose.

controllers/backstage_controller.go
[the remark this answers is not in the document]
jianrongzhang89: They are different objects but are related to the local psql db deployment.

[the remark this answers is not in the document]
gazarenkov: yes there is a logical relationship, but local db service is also logically related to backstage pod, so what's the conclusion? :) 
I mean I still do not understand your original question, maybe you can share some more context

config/manager/default-config/db-statefulset.yaml
gazarenkov: I guess it is not, it works well for me. Could you elaborate?

api/v1alpha1/backstage_types.go
```diff
@@ -266,3 +258,7 @@ type TLS struct {
 func init() {
 	SchemeBuilder.Register(&Backstage{}, &BackstageList{})
 }
+
+func (s BackstageSpec) LocalDbEnabled() bool {
```
rm3l, line 262: nit: can we rename this into `IsLocalDbEnabled`? I think making functions returning a boolean sound like a yes/no question makes the intent clearer.

pkg/model/db-statefulset.go
```diff
@@ -0,0 +1,108 @@
...
+
+	} else {
+		b.statefulSet.Spec.Template.Spec.Containers[0].EnvFrom = append(b.statefulSet.Spec.Template.Spec.Containers[0].EnvFrom, envFrom)
+	}
+	b.secretName = envFrom.SecretRef.Name
+}
+
+// returns DB container
+func (b *DbStatefulSet) container() *corev1.Container {
+	return &b.podSpec().Containers[0]
```
rm3l, line 102: It seems this list of containers might be empty. I tried to apply the [`examples/janus-cr-with-app-configs.yaml` CF](https://github.com/janus-idp/operator/blob/main/examples/janus-cr-with-app-configs.yaml) (after renaming `spec.enableLocalDb` into `spec.database.enabledLocalDb` to fix the CR validation error), and the controller keeps crashing while doing the reconciliation:

```
2024-01-10T16:19:19Z    INFO    Observed a panic in reconciler: runtime error: index out of range [0] with length 0     {"controller": "backstage", "controllerGroup": "janus-idp.io", "control
lerKind": "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "76864ec0
-6758-4d23-93a2-c5036c133d7f"}
panic: runtime error: index out of range [0] with length 0 [recovered]
        panic: runtime error: index out of range [0] with length 0

goroutine 163 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:115 +0x1fa
panic({0x18125a0, 0xc000fcc690})
        /usr/lib/golang/src/runtime/panic.go:884 +0x213
janus-idp.io/backstage-operator/pkg/model.(*DbStatefulSet).container(...)
        /workspace/pkg/model/db-statefulset.go:102
janus-idp.io/backstage-operator/pkg/model.(*DbStatefulSet).Object(0xc0014a00c0)
        /workspace/pkg/model/db-statefulset.go:50 +0xe6
janus-idp.io/backstage-operator/pkg/model.InitObjects({_, _}, {{{0x15ecb9a, 0x9}, {0xc00138c180, 0x15}}, {{0xc0000d6b20, 0x20}, {0x0, 0x0}, ...}, ...}, ...)
        /workspace/pkg/model/runtime.go:94 +0x1c4
janus-idp.io/backstage-operator/controllers.(*BackstageReconciler).Reconcile(0xc000539980, {0x1b7b180, 0xc000391cb0}, {{{0xc0004f42d6, 0x5}, {0xc0000d6b20, 0x20}}})
        /workspace/controllers/backstage_controller.go:109 +0x3f1
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0x1b7b180?, {0x1b7b180?, 0xc000391cb0?}, {{{0xc0004f42d6?, 0x16a1600?}, {0xc0000d6b20?, 0x1b68bd8?}}})
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:118 +0xc8
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc000468320, {0x1b7b0d8, 0xc0001186e0}, {0x1798100?, 0xc0002fd220?})
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:314 +0x377
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc000468320, {0x1b7b0d8, 0xc0001186e0})
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265 +0x1d9
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226 +0x85
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:222 +0x587
```

<details>
<summary>Full logs</summary>

```
$ kubectl logs -n openshift-operators backstage-controller-manager-6b9f4bc857-vc89c                                                                                                            
Alias tip: kl -n openshift-operators backstage-controller-manager-6b9f4bc857-vc89c                                                                                                             
2024-01-10T16:18:51Z    INFO    controller-runtime.metrics      Metrics server is starting to listen    {"addr": "127.0.0.1:8080"}                                                             
2024-01-10T16:18:51Z    INFO    setup   starting manager with parameters:       {"own-runtime": true, "env.LOCALBIN": "", "isOpenShift": true}                                                 
2024-01-10T16:18:51Z    INFO    starting server {"path": "/metrics", "kind": "metrics", "addr": "127.0.0.1:8080"}                                                                              
2024-01-10T16:18:51Z    INFO    Starting server {"kind": "health probe", "addr": "[::]:8081"}                                                                                                  
I0110 16:18:51.338651       1 leaderelection.go:250] attempting to acquire leader lease openshift-operators/06bdbdd5.janus-idp.io...                                                           
I0110 16:19:08.268597       1 leaderelection.go:260] successfully acquired lease openshift-operators/06bdbdd5.janus-idp.io                                                                     
2024-01-10T16:19:08Z    DEBUG   events  backstage-controller-manager-6b9f4bc857-vc89c_a06f486e-5aa0-4493-a2c7-c4711d31ae2a became leader        {"type": "Normal", "object": {"kind":"Lease","n
amespace":"openshift-operators","name":"06bdbdd5.janus-idp.io","uid":"30dea8a2-061c-4899-a575-87d44c191ebf","apiVersion":"coordination.k8s.io/v1","resourceVersion":"11933"}, "reason": "Leader
Election"}                                                                                                                                                                                     
2024-01-10T16:19:08Z    INFO    Starting EventSource    {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "source": "kind source: *v1alpha1.Backsta
ge"}                                                                                                                                                                                           
2024-01-10T16:19:08Z    INFO    Starting EventSource    {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "source": "kind source: *v1.Deployment"} 
2024-01-10T16:19:08Z    INFO    Starting EventSource    {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "source": "kind source: *v1.Service"}    
2024-01-10T16:19:08Z    INFO    Starting EventSource    {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "source": "kind source: *v1.StatefulSet"}
2024-01-10T16:19:08Z    INFO    Starting Controller     {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage"}                                          
2024-01-10T16:19:08Z    INFO    Starting workers        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "worker count": 1}                       
2024-01-10T16:19:14Z    DEBUG   starting reconciliation (namespace: "my-ns/my-backstage-app-with-app-config")   {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind"
: "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "d592d4ea-c0ef-40
79-9d53-80c2b86f5e4a"}                                                                                                                                                                         
2024-01-10T16:19:14Z    ERROR   Reconciler error        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "Backstage": {"name":"my-backstage-app-wi
th-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "d592d4ea-c0ef-4079-9d53-80c2b86f5e4a", "error": "failed to preprocess ba
ckstage spec: failed to get configMap my-backstage-config-backend-auth: ConfigMap \"my-backstage-config-backend-auth\" not found"}                                                             
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler                                                                                                          
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:324
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226
2024-01-10T16:19:14Z    DEBUG   starting reconciliation (namespace: "my-ns/my-backstage-app-with-app-config")   {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind"
: "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "60a9410a-a75c-4f
3c-9d6c-93b704afe493"}
2024-01-10T16:19:14Z    ERROR   Reconciler error        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "Backstage": {"name":"my-backstage-app-wi
th-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "60a9410a-a75c-4f3c-9d6c-93b704afe493", "error": "failed to preprocess ba
ckstage spec: failed to get configMap my-backstage-config-backend-auth: ConfigMap \"my-backstage-config-backend-auth\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:324
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226
2024-01-10T16:19:14Z    DEBUG   starting reconciliation (namespace: "my-ns/my-backstage-app-with-app-config")   {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind"
: "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "80fe151b-137d-4a
5f-b684-7572d158ce82"}
2024-01-10T16:19:14Z    ERROR   Reconciler error        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "Backstage": {"name":"my-backstage-app-wi
th-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "80fe151b-137d-4a5f-b684-7572d158ce82", "error": "failed to preprocess ba
ckstage spec: failed to get configMap my-backstage-config-backend-auth: ConfigMap \"my-backstage-config-backend-auth\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:324
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226
2024-01-10T16:19:14Z    DEBUG   starting reconciliation (namespace: "my-ns/my-backstage-app-with-app-config")   {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind"
: "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "15e036c6-c2e5-4f
0b-8c4d-083ba0aca9ed"}
2024-01-10T16:19:14Z    ERROR   Reconciler error        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "Backstage": {"name":"my-backstage-app-wi
th-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "15e036c6-c2e5-4f0b-8c4d-083ba0aca9ed", "error": "failed to preprocess ba
ckstage spec: failed to get configMap my-backstage-config-backend-auth: ConfigMap \"my-backstage-config-backend-auth\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:324
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2                                                                                          17:19:34 [78/55999]
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226
2024-01-10T16:19:14Z    DEBUG   starting reconciliation (namespace: "my-ns/my-backstage-app-with-app-config")   {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind"
: "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "bedb5b6c-3206-46
c8-bc07-9674d54791b0"}
2024-01-10T16:19:14Z    ERROR   Reconciler error        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "Backstage": {"name":"my-backstage-app-wi
th-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "bedb5b6c-3206-46c8-bc07-9674d54791b0", "error": "failed to preprocess ba
ckstage spec: failed to get configMap my-backstage-config-backend-auth: ConfigMap \"my-backstage-config-backend-auth\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:324
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226
2024-01-10T16:19:14Z    DEBUG   starting reconciliation (namespace: "my-ns/my-backstage-app-with-app-config")   {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind"
: "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "94bfb2a0-d7b9-44
a6-949d-12956fc814fc"}
2024-01-10T16:19:14Z    ERROR   Reconciler error        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "Backstage": {"name":"my-backstage-app-wi
th-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "94bfb2a0-d7b9-44a6-949d-12956fc814fc", "error": "failed to preprocess ba
ckstage spec: failed to get configMap my-backstage-config-backend-auth: ConfigMap \"my-backstage-config-backend-auth\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:324
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226
2024-01-10T16:19:15Z    DEBUG   starting reconciliation (namespace: "my-ns/my-backstage-app-with-app-config")   {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind"
: "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "1879853c-5152-4e
24-85be-9349a2f025b5"}
2024-01-10T16:19:15Z    ERROR   Reconciler error        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "Backstage": {"name":"my-backstage-app-wi
th-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "1879853c-5152-4e24-85be-9349a2f025b5", "error": "failed to preprocess ba
ckstage spec: failed to get configMap my-backstage-config-cm1: ConfigMap \"my-backstage-config-cm1\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:324
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226
2024-01-10T16:19:15Z    DEBUG   starting reconciliation (namespace: "my-ns/my-backstage-app-with-app-config")   {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind"
: "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "5de86128-5e59-4f
88-af1b-b47a4a58ffe8"}
2024-01-10T16:19:15Z    ERROR   Reconciler error        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "Backstage": {"name":"my-backstage-app-wi
th-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "5de86128-5e59-4f88-af1b-b47a4a58ffe8", "error": "failed to preprocess ba
ckstage spec: failed to get configMap my-backstage-config-cm1: ConfigMap \"my-backstage-config-cm1\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:324
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226
2024-01-10T16:19:16Z    DEBUG   starting reconciliation (namespace: "my-ns/my-backstage-app-with-app-config")   {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind"
: "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "15f2ccc4-dccb-4f
e2-a121-5f09af935046"}
2024-01-10T16:19:16Z    ERROR   Reconciler error        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "Backstage": {"name":"my-backstage-app-wi
th-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "15f2ccc4-dccb-4fe2-a121-5f09af935046", "error": "failed to preprocess ba
ckstage spec: failed to get ConfigMap my-backstage-extra-files-cm1: ConfigMap \"my-backstage-extra-files-cm1\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:324
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226
2024-01-10T16:19:17Z    DEBUG   starting reconciliation (namespace: "my-ns/my-backstage-app-with-app-config")   {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind"
: "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "441237ee-15e3-41
bf-90f8-47c0577a575e"}
2024-01-10T16:19:17Z    ERROR   Reconciler error        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "Backstage": {"name":"my-backstage-app-wi
th-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "441237ee-15e3-41bf-90f8-47c0577a575e", "error": "failed to preprocess ba
ckstage spec: failed to get ConfigMap my-backstage-extra-files-cm1: ConfigMap \"my-backstage-extra-files-cm1\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:324
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226
2024-01-10T16:19:19Z    DEBUG   starting reconciliation (namespace: "my-ns/my-backstage-app-with-app-config")   {"controller": "backstage", "controllerGroup": "janus-idp.io"17:19:34 [4/55999]
: "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "76864ec0-6758-4d
23-93a2-c5036c133d7f"}
2024-01-10T16:19:19Z    DEBUG   failed to read default value for optional key. Ignored 
        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace"
: "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "76864ec0-6758-4d23-93a2-c5036c133d7f", "configmap-envs.yaml": "failed to read default value for the key configmap-envs.y
aml, reason: failed to read YAML file: open default-config/configmap-envs.yaml: no such file or directory"}
2024-01-10T16:19:19Z    DEBUG   failed to read default value for optional key. Ignored 
        {"controller": "backstage", "controllerGroup": "janus-idp.io", "controllerKind": "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace"
: "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "76864ec0-6758-4d23-93a2-c5036c133d7f", "configmap-files.yaml": "failed to read default value for the key configmap-files
.yaml, reason: failed to read YAML file: open default-config/configmap-files.yaml: no such file or directory"}
2024-01-10T16:19:19Z    INFO    Observed a panic in reconciler: runtime error: index out of range [0] with length 0     {"controller": "backstage", "controllerGroup": "janus-idp.io", "control
lerKind": "Backstage", "Backstage": {"name":"my-backstage-app-with-app-config","namespace":"my-ns"}, "namespace": "my-ns", "name": "my-backstage-app-with-app-config", "reconcileID": "76864ec0
-6758-4d23-93a2-c5036c133d7f"}
panic: runtime error: index out of range [0] with length 0 [recovered]
        panic: runtime error: index out of range [0] with length 0

goroutine 163 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:115 +0x1fa
panic({0x18125a0, 0xc000fcc690})
        /usr/lib/golang/src/runtime/panic.go:884 +0x213
janus-idp.io/backstage-operator/pkg/model.(*DbStatefulSet).container(...)
        /workspace/pkg/model/db-statefulset.go:102
janus-idp.io/backstage-operator/pkg/model.(*DbStatefulSet).Object(0xc0014a00c0)
        /workspace/pkg/model/db-statefulset.go:50 +0xe6
janus-idp.io/backstage-operator/pkg/model.InitObjects({_, _}, {{{0x15ecb9a, 0x9}, {0xc00138c180, 0x15}}, {{0xc0000d6b20, 0x20}, {0x0, 0x0}, ...}, ...}, ...)
        /workspace/pkg/model/runtime.go:94 +0x1c4
janus-idp.io/backstage-operator/controllers.(*BackstageReconciler).Reconcile(0xc000539980, {0x1b7b180, 0xc000391cb0}, {{{0xc0004f42d6, 0x5}, {0xc0000d6b20, 0x20}}})
        /workspace/controllers/backstage_controller.go:109 +0x3f1
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0x1b7b180?, {0x1b7b180?, 0xc000391cb0?}, {{{0xc0004f42d6?, 0x16a1600?}, {0xc0000d6b20?, 0x1b68bd8?}}})
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:118 +0xc8
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc000468320, {0x1b7b0d8, 0xc0001186e0}, {0x1798100?, 0xc0002fd220?})
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:314 +0x377
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc000468320, {0x1b7b0d8, 0xc0001186e0})
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265 +0x1d9
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226 +0x85
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:222 +0x587
```

</details>

config/manager/kustomization.yaml
```diff
@@ -19,6 +22,7 @@ configMapGenerator:
   - default-config/db-service.yaml
   - default-config/db-service-hl.yaml
   - default-config/db-secret.yaml
-  - default-config/backend-auth-configmap.yaml
-  - default-config/dynamic-plugins-configmap.yaml
+  - default-config/dynamic-plugins.yaml
+  - default-config/app-config.yaml
```
rm3l, line 26: nit: can you sort this list alphabetically? I think it would help the reviewers.

pkg/model/db-statefulset.go
rm3l: Hmm, actually it also happened with the minimal `bs1.yaml` CR: https://github.com/janus-idp/operator/blob/main/examples/bs1.yaml
Am I missing something?

main.go
```diff
@@ -104,12 +104,10 @@ func main() {
 	}
 
 	if err = (&controller.BackstageReconciler{
-		Client:         mgr.GetClient(),
-		Scheme:         mgr.GetScheme(),
-		OwnsRuntime:    ownRuntime,
-		IsOpenShift:    isOpenShift,
-		PsqlImage:      os.Getenv(backstageiov1alpha1.EnvPostGresImage),
-		BackstageImage: os.Getenv(backstageiov1alpha1.EnvBackstageImage),
```
rm3l, line 112: I guess you can also remove both fields from the `BackstageReconciler` struct?

pkg/model/runtime.go
```diff
@@ -0,0 +1,201 @@
...
+				overlayErr = fmt.Errorf("failed to read overlay value for the key %s, reason: %s", conf.Key, err)
+			}
+		}
+
+		// throw the error if raw configuration exists and is invalid
+		// throw the error if there is invalid or no configuration (default|raw) for Mandatory object
+		// continue if there is invalid or no configuration (default|raw) for Optional object
+		// TODO separate the case when configuration does not exist (intentionally) from invalid configuration
+		if overlayErr != nil || (!overlayExist && defaultErr != nil) {
+			if conf.need == Mandatory || (conf.need == ForLocalDatabase && *backstageSpec.Database.EnableLocalDb) {
```
rm3l, line 113: Just to be safe, could we use the `LocalDbEnabled()` method you introduced instead?
```suggestion
			if conf.need == Mandatory || (conf.need == ForLocalDatabase && backstageSpec.LocalDbEnabled()) {
```

pkg/model/db-statefulset.go
rm3l: To clarify, this happened after building the container images and running the controller in the cluster (`make release-build release-push catalog-update`).
Now when running locally with `make install run`, I'm seeing a different error: it seems `<RELATED_IMAGE_backstage>` and `<RELATED_IMAGE_postgresql>` are not being replaced (I didn't set those env vars before running the operator):
```
$ kubectl describe pod bs1-db-statefulset-0
[...]
Events:
  Type     Reason                  Age                     From                     Message
  ----     ------                  ----                    ----                     -------
  Normal   Scheduled               5m40s                   default-scheduler        Successfully assigned my-ns/bs1-db-statefulset-0 to ci-ln-ywvkv12-72292-56c76-worker-b-lb5r2
  Normal   SuccessfulAttachVolume  5m36s                   attachdetach-controller  AttachVolume.Attach succeeded for volume "pvc-07468e13-d1c3-456e-8a9d-800ee7ec1bf1"
  Normal   AddedInterface          5m32s                   multus                   Add eth0 [10.128.2.16/23] from ovn-kubernetes
  Warning  Failed                  3m33s (x12 over 5m32s)  kubelet                  Error: InvalidImageName
  Warning  InspectFailed           31s (x26 over 5m32s)    kubelet                  Failed to apply default image tag "<RELATED_IMAGE_postgresql>": couldn't parse image reference "<RELATED_IMAGE_postgresql>": invalid reference format: repository name must be lowercase
```

pkg/model/runtime.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+		}
+
+		// do not add if ForLocalDatabase and LocalDb is disabled
+		if !backstageSpec.LocalDbEnabled() && conf.need == ForLocalDatabase {
+			continue
+		}
+
+		// do not add if ForOpenshift and cluster is not Openshift
+		if !isOpenshift && conf.need == ForOpenshift {
```
rm3l, line 127: I think we should also check if route is explicitly disabled by the user in the CR spec, and not add it if so.

pkg/model/db-statefulset.go
```diff
@@ -0,0 +1,108 @@
...
+// Local DB pod considered to have single container
+func (b *DbStatefulSet) setSecretNameEnvFrom(envFrom corev1.EnvFromSource) {
+
+	// it is possible that Secret name already set by default configuration
+	// has to be overriden in this case
+	if b.secretName != "" {
+		var ind int
+		for i, v := range b.container().EnvFrom {
+			if v.SecretRef.Name == b.secretName {
+				ind = i
```
rm3l, line 88: Do we need that additional `ind` variable? Why not do the change right away and break?
```
b.statefulSet.Spec.Template.Spec.Containers[0].EnvFrom[i] = envFrom
```

api/v1alpha1/backstage_types.go
gazarenkov: +1

config/manager/kustomization.yaml
gazarenkov: +1

pkg/model/runtime.go
gazarenkov: I think so. +1

pkg/model/runtime.go
gazarenkov: Ok but not here, it is for Phase 3: process Backstage.spec, getting final desired state...
For my understanding - what would be the reason of disabling it per CR?

pkg/model/db-statefulset.go
gazarenkov: +1 
... overcomplexity as a result of evolution  :)

main.go
gazarenkov: ah, sure, I though I did

pkg/model/runtime.go
rm3l: > Ok but not here, it is for Phase 3: process Backstage.spec, getting final desired state... 

Okay, that makes sense.

> For my understanding - what would be the reason of disabling it per CR?

When we demo'ed the automatic Route creation, Tomas mentioned that we should also provide a way to disable it in the CR (https://github.com/janus-idp/operator/issues/93). Users might not want to have a Route, if they want to access their Backstage instance differently.
As a general rule of thumb, we should be careful with things that we create automatically, and should also provide a way to disable that behavior in the CR.

rm3l (changes requested):
I applied the simplest CR (`examples/bs1.yaml`) manifest. By default, on OpenShift, a Route is created (as expected), but it does not allow reaching out the Pod:

image
image


Looking at the Route created:
```
$ oc get route bs1-route -o yaml                                                                                                                
[...]
spec:
  host: bs1-route-my-ns.apps.rosa.qykoa-p9t92-h6k.u86b.p3.openshiftapps.com
  path: /
  port:
    targetPort: http-backend
  tls:
    insecureEdgeTerminationPolicy: Redirect
    termination: edge
  to:
    kind: Service
    name: bs1-route
    weight: 100
  wildcardPolicy: None
```

It looks like the `to` Service in the Route spec is not correct. It should be `bs1-service` instead in this case.

rm3l (commented):
It seems `spec.application.dynamicPluginsConfigMapName` field is not taken into account properly, so users cannot configure their list of dynamic plugins. I could only see the default ones loaded.
You can reproduce by applying the [`examples/janus-cr-with-app-configs.yaml`](https://github.com/janus-idp/operator/blob/main/examples/janus-cr-with-app-configs.yaml) CR (which is supposed to load 3 additional dynamic plugins: https://github.com/janus-idp/operator/blob/main/examples/janus-cr-with-app-configs.yaml#L110-L132) and `GET`'ing the `/api/dynamic-plugins-info/loaded-plugins` API endpoint of Backstage (or looking at the Backstage init container logs).

- On the `main` branch:

```json
[
  {
    "name": "backstage-plugin-catalog-backend-module-github-dynamic",
    "version": "0.4.4",
    "platform": "node",
    "role": "backend-plugin-module"
  },
  {
    "name": "@dfatwork-pkgs/explore-backend-wrapped-dynamic",
    "version": "0.0.9-next.11",
    "platform": "node",
    "role": "backend-plugin"
  },
  {
    "name": "@dfatwork-pkgs/scaffolder-backend-module-http-request-wrapped-dynamic",
    "version": "4.0.9-0",
    "platform": "node",
    "role": "backend-plugin-module"
  },
  {
    "name": "@janus-idp/backstage-scaffolder-backend-module-quay-dynamic",
    "version": "1.2.4",
    "platform": "node",
    "role": "backend-plugin-module"
  },
  {
    "name": "@janus-idp/backstage-scaffolder-backend-module-regex-dynamic",
    "version": "1.2.4",
    "platform": "node",
    "role": "backend-plugin-module"
  },
  {
    "name": "roadiehq-scaffolder-backend-module-utils-dynamic",
    "version": "1.10.4",
    "platform": "node",
    "role": "backend-plugin-module"
  }
]
```

- But with this PR, I can only see the default ones:
```json
[
  {
    "name": "@janus-idp/backstage-scaffolder-backend-module-quay-dynamic",
    "version": "1.2.4",
    "platform": "node",
    "role": "backend-plugin-module"
  },
  {
    "name": "@janus-idp/backstage-scaffolder-backend-module-regex-dynamic",
    "version": "1.2.4",
    "platform": "node",
    "role": "backend-plugin-module"
  },
  {
    "name": "roadiehq-scaffolder-backend-module-utils-dynamic",
    "version": "1.10.4",
    "platform": "node",
    "role": "backend-plugin-module"
  }
]
```

This means to me that only the default config for dynamic plugins has been loaded, which can be confirmed by looking at the `dynamic-plugins-root/app-config.dynamic-plugins.yaml` (generated by the init container):

- On `main`:
```
bash-5.1$ cat dynamic-plugins-root/app-config.dynamic-plugins.yaml 
catalog:
  providers:
    github:
      myorg:
        organization: ${GITHUB_ORG}
        schedule:
          frequency:
            minutes: 30
          initialDelay:
            seconds: 15
          timeout:
            minutes: 3
dynamicPlugins:
  rootDirectory: dynamic-plugins-root
proxy:
  endpoints:
    /explore-backend-completed:
      target: http://localhost:7017
```

- With this PR:
```
bash-5.1$ cat dynamic-plugins-root/app-config.dynamic-plugins.yaml 
dynamicPlugins:
  rootDirectory: dynamic-plugins-root
```

pkg/model/runtime.go
gazarenkov: fixed as suggested for the time

pkg/model/db-statefulset.go
gazarenkov: is not it gone after last fixes?

gazarenkov:
> I applied the simplest CR (`examples/bs1.yaml`) manifest. By default, on OpenShift, a Route is created (as expected), but it does not allow reaching out the Pod:
> 
> image image
> 
> Looking at the Route created:
> 
> ```
> $ oc get route bs1-route -o yaml                                                                                                                
> [...]
> spec:
>   host: bs1-route-my-ns.apps.rosa.qykoa-p9t92-h6k.u86b.p3.openshiftapps.com
>   path: /
>   port:
>     targetPort: http-backend
>   tls:
>     insecureEdgeTerminationPolicy: Redirect
>     termination: edge
>   to:
>     kind: Service
>     name: bs1-route
>     weight: 100
>   wildcardPolicy: None
> ```
> 
> It looks like the `to` Service in the Route spec is not correct. It should be `bs1-service` instead in this case.

must be fixed, please check

pkg/model/db-statefulset.go
rm3l: Nope, I am still seeing a panic. It seems to happen only when `RELATED_IMAGE_*` env var are set.
You can reproduce locally with something like:
```sh
$ RELATED_IMAGE_backstage=busybox make install run
```

Then you'll see the panic when reconciling any CR.

gazarenkov: It should be OK now

Running as you suggested:
Error: failed to start container "backstage-backend": Error response from daemon: failed to create shim task: OCI runtime create failed: runc  │
│ create failed: unable to start container process: exec: "--config": executable file not found in $PATH: unknown 

But it is expected I think as the image is just busybox

rm3l: Nice, with the latest changes, I'm no longer facing a panic.
But I noticed one more thing to change on the Backstage Pod.
The image should be replaced on **all** containers of the Pod Spec, including the init container. Currently, it is only replaced for the main container. Init container is still using the default `quay.io/janus-idp/backstage-showcase:next` image, which might be problematic for downstream builds or disconnected environments.
Example if I set `RELATED_IMAGE_backstage` to `quay.io/rhdh/rhdh-hub-rhel9:1.0-200` in the controller environment:
```
$ kubectl get deployment bs1-deployment -o yaml
[...]
spec:
 template:
    spec:
      containers:
        image: quay.io/rhdh/rhdh-hub-rhel9:1.0-200
        imagePullPolicy: IfNotPresent
        name: backstage-backend
      initContainers:
        image: quay.io/janus-idp/backstage-showcase:next
        imagePullPolicy: IfNotPresent
        name: install-dynamic-plugins
[...]
```

rm3l:
> > I applied the simplest CR (`examples/bs1.yaml`) manifest. By default, on OpenShift, a Route is created (as expected), but it does not allow reaching out the Pod:
> > image image
> > Looking at the Route created:
> > ```
> > $ oc get route bs1-route -o yaml                                                                                                                
> > [...]
> > spec:
> >   host: bs1-route-my-ns.apps.rosa.qykoa-p9t92-h6k.u86b.p3.openshiftapps.com
> >   path: /
> >   port:
> >     targetPort: http-backend
> >   tls:
> >     insecureEdgeTerminationPolicy: Redirect
> >     termination: edge
> >   to:
> >     kind: Service
> >     name: bs1-route
> >     weight: 100
> >   wildcardPolicy: None
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It looks like the `to` Service in the Route spec is not correct. It should be `bs1-service` instead in this case.
> 
> must be fixed, please check

Route is now working properly with the latest changes. Thanks!

pkg/model/db-statefulset.go
rm3l: Same behavior when overriding the image from the one set in `spec.application.image`. It should override the image on **all** containers of the Pod Spec, including the init container. Currently, it only overrides the main container image.

rm3l: FYI, we had a `visitContainers` utility function to consistently perform some operations on a `PodTemplateSpec`: https://github.com/janus-idp/operator/blob/bac627410197f4b1d6a120b55189d7464181fa3c/controllers/backstage_deployment.go#L136-L147

I think it can help prevent such issues in the future.

jianrongzhang89:
Please fix the merge conflicts.

gazarenkov:
> /api/dynamic-plugins-info/loaded-plugins

Looks like it works

pkg/model/route.go
```diff
@@ -0,0 +1,64 @@
...
+func (b *BackstageRoute) Object() client.Object {
+	return b.route
+}
+
+// implementation of BackstageObject interface
+func (b *BackstageRoute) EmptyObject() client.Object {
+	return &openshift.Route{}
+}
+
+// implementation of BackstageObject interface
```
jianrongzhang89, line 48: Why the route related features are dropped?
https://github.com/janus-idp/operator/pull/94
https://github.com/janus-idp/operator/pull/79

Refactoring should not remove existing capabilities.

jianrongzhang89 (changes requested):

controllers/backstage_controller.go
```diff
@@ -147,144 +133,46 @@ func (r *BackstageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-	if len(r.BackstageImage) == 0 {
-		r.BackstageImage = "quay.io/janus-idp/backstage-showcase:next"
-		log.Info("Enviroment variable is not set, default is used", bs.EnvBackstageImage, r.BackstageImage)
-	}
-
 	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&bs.Backstage{})
 
 	if r.OwnsRuntime {
 		builder.Owns(&appsv1.Deployment{}).
```
jianrongzhang89, line 173: My concern is that this will cause the operator to watch ALL deployment, service and statefulset objects in the entire cluster. This can lead to dramatic resource consumptions in a production cluster where there are large number of such clusters. I think we should drop this until we are able to do a load/performance test to evaluate the performance impact.

pkg/model/route.go
gazarenkov: It is not dropped, tests passed and as mentioned https://github.com/janus-idp/operator/pull/70#issuecomment-1887454951 "Route is now working properly with the latest changes"

controllers/backstage_controller.go
gazarenkov: OK for me, it was kinda experimental thing not related to this PR (you can see something like this in the main branch too ;) )

pkg/model/route.go
jianrongzhang89: @gazarenkov The route only works in the default configuration. I tested the following CR and found that:
1) support for route.enabled is missing.
2) Support for route.host/subdomain does not work.
3)Support for route.tls certificates in the route does not work.
So basically all functionalities implemented in the previous PRs are dropped. 
https://github.com/janus-idp/operator/pull/94
https://github.com/janus-idp/operator/pull/79

```
apiVersion: janus-idp.io/v1alpha1
kind: Backstage
metadata:
  name: bs1
spec:
  database:
    enableLocalDb: true
    authSecretName: test-postgres-secrets
  application:
    imagePullSecrets:
    - my-pull-secret
    replicas: 2
    route:
      enabled: false
      host: test.com
      subdomain: mysubdomain
    extraEnvs:
      envs:
        - name: myenv1
          value: my-value-1
        - name: myenv2
          value: my-value-2
```

jianrongzhang89:
@gazarenkov can you please add in the PR description with the list of all test cases that were performed in this PR as well as the expected result for each one? We need this list to verify the PR.

jianrongzhang89:
@gazarenkov also noticed that the previous work for CR changes got lost. I did the following test:
1) apply examples/bs1.yaml to deploy backstage CR bs1.
2) Apply the changes for the CR:
```
apiVersion: janus-idp.io/v1alpha1
kind: Backstage
metadata:
  name: bs1
spec:
  database:
    enableLocalDb: false
    authSecretName: test-postgres-secrets
  application:
    replicas: 2
    imagePullSecrets:
    - my-pull-secret
    extraEnvs:
      envs:
        - name: myenv1
          value: my-value-1
        - name: myenv2
          value: my-value-2
``` 
3) I would expect the following to happen:
a) the local db gets deleted.
b) backstage deployment to have replicas of 2.
c) imagePullSecrets has been set to my-pull-secret
d) and env vars myenv1 and myenv2 to appear in the deployement.
However none of these actually happened.

rm3l (commented):
@gazarenkov I tested the latest commit ([`2dceb0d` (#70)](https://github.com/janus-idp/operator/pull/70/commits/2dceb0de3746bad16b70be8a382d6f0fa19eebe1)), and it looks like the `route.enabled` is still ignored (https://github.com/janus-idp/operator/pull/70#discussion_r1453533167), and CR updates are not working as expected (https://github.com/janus-idp/operator/pull/70#issuecomment-1894804139).

Also, it looks like the Git conflicts are still there after your latest update..

jianrongzhang89:
@gazarenkov Now I got an error when deploying examples/bs1.yaml. Can you take a look?
```
status:
  conditions:
  - lastTransitionTime: "2024-01-25T16:52:07Z"
    message: 'failed to apply backstage objects failed to update object bs1-route:
      routes.route.openshift.io "bs1-route" is invalid: metadata.resourceVersion:
      Invalid value: 0x0: must be specified for an update'
    reason: DeployFailed
    status: "False"
    type: Deployed
```

jianrongzhang89:
@gazarenkov I found that only one service is deployed for the local db. The db statefulset requires a headless service (with serviceName in the statefulset spec) but the headless service is missing. Can you double check?

controllers/backstage_controller.go
```diff
@@ -103,219 +104,140 @@ func (r *BackstageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			lg.V(1).Info("Create object ", "obj", obj.Object().GetName())
+			continue
 		}
-	} else {
-		lg.V(1).Info("custom configuration configMap exists but no such key, applying default config", "configMap", cm.Name, "key", key)
-		err := readYaml([]byte(val), object)
-		if err != nil {
-			return fmt.Errorf("failed to read YAML: %w", err)
+
+		if err := r.Update(ctx, obj.Object()); err != nil {
```
jianrongzhang89, line 177: Simply calling update() will cause unexpected result because the existing k8s object's meta data information (such as labels, annotations) may get lost. Administrators often may want to add additional labels or annotations in order to manager the cluster resources.
You should consider to use the CreateOrUpdate() library call used currently, or you may have to implement the logic to merge the configuration before calling update.

gazarenkov: As discussed, if we want Backstage owning the objects (which is the case) the only way to reshape runtime objects is to change default/raw/spec configuration, that is the desired state

gazarenkov:
> @gazarenkov I found that only one service is deployed for the local db. The db statefulset requires a headless service (with serviceName in the statefulset spec) but the headless service is missing. Can you double check?

If you consider it important please make sure it is added in test suite, it is not that obvious w/o ability to test it.

gazarenkov:
> @gazarenkov Now I got an error when deploying examples/bs1.yaml. Can you take a look?
> 
> ```
> status:
>   conditions:
>   - lastTransitionTime: "2024-01-25T16:52:07Z"
>     message: 'failed to apply backstage objects failed to update object bs1-route:
>       routes.route.openshift.io "bs1-route" is invalid: metadata.resourceVersion:
>       Invalid value: 0x0: must be specified for an update'
>     reason: DeployFailed
>     status: "False"
>     type: Deployed
> ```

Do you mean you 

> @gazarenkov Now I got an error when deploying examples/bs1.yaml. Can you take a look?
> 
> ```
> status:
>   conditions:
>   - lastTransitionTime: "2024-01-25T16:52:07Z"
>     message: 'failed to apply backstage objects failed to update object bs1-route:
>       routes.route.openshift.io "bs1-route" is invalid: metadata.resourceVersion:
>       Invalid value: 0x0: must be specified for an update'
>     reason: DeployFailed
>     status: "False"
>     type: Deployed
> ```

Do you mean this failed for you?
https://github.com/janus-idp/operator/blob/main/examples/bs1.yaml

rm3l:
> > @gazarenkov Now I got an error when deploying examples/bs1.yaml. Can you take a look?
> > ```
> > status:
> >   conditions:
> >   - lastTransitionTime: "2024-01-25T16:52:07Z"
> >     message: 'failed to apply backstage objects failed to update object bs1-route:
> >       routes.route.openshift.io "bs1-route" is invalid: metadata.resourceVersion:
> >       Invalid value: 0x0: must be specified for an update'
> >     reason: DeployFailed
> >     status: "False"
> >     type: Deployed
> > ```
> 
> Do you mean this failed for you? [`main`/examples/bs1.yaml](https://github.com/janus-idp/operator/blob/main/examples/bs1.yaml?rgh-link-date=2024-01-26T11%3A27%3A23Z)

Got the same issue after applying `bs1.yaml`. Note that all the resources (including the route) are created and work as expected. Only the CR status contains this error message, which could be misleading I guess..

controllers/backstage_controller.go
jianrongzhang89: Yes, we do want to have runtime objects to be based on our configuration, however we should not prevent admins to add the labels/annotations to them, and this is often done as automation scripts. k8s may inject additional information to the metadata.  This will be a serious limitation which will not get accepted by the field.

config/manager/default-config/secret-envs.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: backend-auth-secret
+stringData:
+  # generated with the command below (from https://janus-idp.io/docs/auth/service-to-service-auth/#setup):
+  # node -p 'require("crypto").randomBytes(24).toString("base64")'
+  BACKEND_SECRET: "R2FxRVNrcmwzYzhhN3l0V1VRcnQ3L1pLT09WaVhDNUEK" # notsecret
```
jianrongzhang89, line 8: We should not use a default credentials here.

controllers/backstage_controller.go
```diff
@@ -103,219 +93,146 @@ func (r *BackstageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	if err != nil {
-		return ctrl.Result{}, err
+		setStatusCondition(&backstage, bs.BackstageConditionTypeDeployed, metav1.ConditionFalse, bs.BackstageConditionReasonFailed, fmt.Sprintf("failed to initialize backstage model %s", err))
+		return ctrl.Result{}, fmt.Errorf("failed to initialize backstage model %w", err)
 	}
 
-	if err := r.reconcileBackstageService(ctx, &backstage, req.Namespace); err != nil {
-		return ctrl.Result{}, err
+	err = r.applyObjects(ctx, bsModel.Objects)
+	if err != nil {
```
jianrongzhang89, line 123: Errors of conflicting objects are temporary. They will be cleared in next reconciliations and should not cause the CR status to get updated.

config/manager/default-config/secret-envs.yaml
gazarenkov: it is a matter of configuration, secret-envs is not mandatory or can be created as configmap-envs (like it managed in the  main branch https://github.com/janus-idp/operator/blob/main/config/manager/default-config/backend-auth-configmap.yaml)

controllers/backstage_controller.go
jianrongzhang89: This kind of conflicting errors are vey common in operators and all they need to do is to retry the reconciliation. They are typically self-recoverable so they should not cause un-necessary status updates.

jianrongzhang89:
> > @gazarenkov I found that only one service is deployed for the local db. The db statefulset requires a headless service (with serviceName in the statefulset spec) but the headless service is missing. Can you double check?
> 
> If you consider it important please make sure it is added in test suite, it is not that obvious w/o ability to test it.

Without such service, it will be confusing for users - the serviceName is set in the statefulset, but the actual service is missing.

This is already in the test cases in main, see https://github.com/janus-idp/operator/blob/main/controllers/backstage_controller_test.go#L396. 
In addition, the helm chart deploys both the regular service and the headless service for the local db too.

Please make sure you do not delete existing test cases during refactoring. Otherwise makes us difficult to validate the changes. IMO, refactoring should not even modify existing test cases so that we know none of existing test cases get broken.

PR #79, opened 2023-12-18
Add Route Host and TLS configuration to Backstage CR

### LINKED ISSUE #13: Specify and implement routing configuration for Backstage server endpoint
The options for exposing Backstage service are:
- Just a Sevice to use for local evaluation/testing 
- Ingress for using in vanilla K8s
- Openshift Route

jianrongzhang89:
## Description
Added host, subdomain and TLS certificates to the OpenShift route configuration in the Backstage CRD.

- Fixes #13 

## PR acceptance criteria

- [X] Tests
- [ ] Documentation

a) Deploy the operator on OpenShift cluster.
b) Deploy CR: examples/bs-route.yaml
c) Check that an OpenShift route with host test-bs-route-<cluster base URL> has been created and the page https:// test-bs-route-<cluster base URL> is accessible.


bundle.Dockerfile
rm3l, line 1: This file was relocated to `docker/bundle.Dockerfile` in #75, I think.

controllers/backstage_route.go
```diff
@@ -56,9 +56,57 @@ func (r *BackstageReconciler) applyBackstageRoute(ctx context.Context, backstage
...
 	}
 	return nil
 }
+
+func (r *BackstageReconciler) applyRouteParamsFromCR(route *openshift.Route, backstage bs.Backstage) {
+	if backstage.Spec.Application == nil || backstage.Spec.Application.Route == nil {
+		return // Nothing to override
+	}
+	routeCfg := backstage.Spec.Application.Route
+	if len(routeCfg.Host) > 0 {
```
rm3l, line 73: nit: this seems more legible to me, as it makes it clear that `routeCfg.Host` is a `string`, not a slice.

```suggestion
	if routeCfg.Host != "" {
```

controllers/backstage_route.go
```diff
@@ -56,9 +56,57 @@ func (r *BackstageReconciler) applyBackstageRoute(ctx context.Context, backstage
...
+		route.Spec.Host = routeCfg.Host
+	}
+	if len(routeCfg.Subdomain) > 0 {
```
rm3l, line 76: Same point.

```suggestion
	if routeCfg.Subdomain != "" {
```

controllers/backstage_route.go
```diff
@@ -56,9 +56,57 @@ func (r *BackstageReconciler) applyBackstageRoute(ctx context.Context, backstage
...
+			Certificate:                   routeCfg.TLS.Certificate,
+			Key:                           routeCfg.TLS.Key,
+			CACertificate:                 routeCfg.TLS.CACertificate,
+			ExternalCertificate: &openshift.LocalObjectReference{
+				Name: routeCfg.TLS.ExternalCertificateSecretName,
+			},
+		}
+		return
+	}
+	if len(routeCfg.TLS.Certificate) > 0 {
```
rm3l, line 95: Same point.

```suggestion
	if routeCfg.TLS.Certificate != "" {
```

controllers/backstage_route.go
```diff
@@ -56,9 +56,57 @@ func (r *BackstageReconciler) applyBackstageRoute(ctx context.Context, backstage
...
+		route.Spec.TLS.Certificate = routeCfg.TLS.Certificate
+	}
+	if len(routeCfg.TLS.Key) > 0 {
```
rm3l, line 98: Same point.

```suggestion
	if routeCfg.TLS.Key != "" {
```

controllers/backstage_route.go
```diff
@@ -56,9 +56,57 @@ func (r *BackstageReconciler) applyBackstageRoute(ctx context.Context, backstage
...
+		route.Spec.TLS.Key = routeCfg.TLS.Key
+	}
+	if len(routeCfg.TLS.Certificate) > 0 {
```
rm3l, line 101: Same point.

```suggestion
	if routeCfg.TLS.Certificate != "" {
```

controllers/backstage_route.go
```diff
@@ -56,9 +56,57 @@ func (r *BackstageReconciler) applyBackstageRoute(ctx context.Context, backstage
...
+		route.Spec.TLS.Certificate = routeCfg.TLS.Certificate
+	}
+	if len(routeCfg.TLS.CACertificate) > 0 {
```
rm3l, line 104: Same point.

```suggestion
	if routeCfg.TLS.CACertificate != "" {
```

controllers/backstage_route.go
```diff
@@ -56,9 +56,57 @@ func (r *BackstageReconciler) applyBackstageRoute(ctx context.Context, backstage
...
+		route.Spec.TLS.CACertificate = routeCfg.TLS.CACertificate
+	}
+	if len(routeCfg.TLS.ExternalCertificateSecretName) > 0 {
```
rm3l, line 107: Same point.

```suggestion
	if routeCfg.TLS.ExternalCertificateSecretName != "" {
```

controllers/backstage_route.go
```diff
@@ -56,9 +56,57 @@ func (r *BackstageReconciler) applyBackstageRoute(ctx context.Context, backstage
...
+		route.Spec.Host = routeCfg.Host
+	}
+	if len(routeCfg.Subdomain) > 0 {
+		route.Spec.Subdomain = routeCfg.Subdomain
+	}
+	if routeCfg.TLS == nil {
+		return
+	}
+	if route.Spec.TLS == nil {
+		route.Spec.TLS = &openshift.TLSConfig{
```
rm3l, line 83: Not sure why all fields (including default ones) are being set here, especially if they are set below again if `route.Spec.TLS != nil`. Would just something like below be sufficient?

```go
if route.Spec.TLS == nil {
  route.Spec.TLS = &openshift.TLSConfig{}
}
// then set the rest of the fields from `routeCfg`
if routeCfg.TLS.Certificate != "" {
  route.Spec.TLS.Certificate = routeCfg.TLS.Certificate
}
// -- Truncated --
```

rm3l (commented):
Works great - thanks! Added a few comments..

controllers/backstage_route.go
rm3l: Ok - fine then.

controllers/backstage_route.go
jianrongzhang89: Actually it is more efficient to use the constructor to initialize the data, and it also avoids the if statement checks later on.

bundle.Dockerfile
jianrongzhang89: bundle.Dockerfile is removed. Unfortunately the implementation is #37 is incomplete (as when the bundle file gets regenerated it will still be in the root directory -- leave it as it for now).

rm3l (approved):

Issue #80, opened 2023-12-19
Enable Operator for restricted network (air-gapped or proxied) environments


rm3l:
As an administrator, I want to run the Operator in a restricted network, or disconnected, environment, So that users can have a fully functional Backstage instance running under such restricted environments.

See https://docs.openshift.com/container-platform/4.14/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-restricted-network_osdk-generating-csvs

See also https://issues.redhat.com/browse/RHIDP-488

nickboldt:
Testing required:
* install with operator
* install with operator using postgres; verify the postgres pod starts and can be used

Doc required:
* how to install in restricted env (steal content from https://access.redhat.com/documentation/en-us/red_hat_openshift_dev_spaces/3.10/html-single/administration_guide/index#installing-devspaces-in-a-restricted-environment)

gazarenkov:
As discussed, manual testing with SCV Airgapped env part of Admin doc are parts of DoD.

rm3l:
> As discussed, manual testing with SCV and Airgapped environment part of Admin doc are parts of DoD.

I've added this point to the documentation issue: https://github.com/janus-idp/operator/issues/28

PR #83, opened 2023-12-19
extract postgres image ref to an env var RELATED_IMAGE_postgresql in CSV (airgap support)

nickboldt:
### What does this PR do?

chore: instead of hardcoding `quay.io/fedora/postgresql-15:latest` in config/manager/default-config/db-statefulset.yaml, extract to an env var RELATED_IMAGE_postgresql = `registry.redhat.io/rhel9/postgresql-15:latest`

Signed-off-by: RHDH Build (rhdh-bot) <rhdh-bot@redhat.com>

### Screenshot/screencast of this PR
N/A

### What issues does this PR fix or reference?
https://github.com/janus-idp/operator/issues/80
https://issues.redhat.com/browse/RHIDP-488

### How to test this PR?
This is an incomplete solution -- it requires an update to the operator itself to fetch `RELATED_IMAGE_postgresql` and use that inside the [config/manager/default-config/db-statefulset.yaml](https://github.com/janus-idp/operator/pull/83/files#diff-b686120f5593397c2cf57779af726c1c9bd968c2f859c946a89d87d4b7c5e49c)

### DIFF
--- a/bundle/manifests/backstage-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/backstage-operator.clusterserviceversion.yaml
@@ -195,6 +195,9 @@ spec:
                 - --health-probe-bind-address=:8081
                 - --metrics-bind-address=127.0.0.1:8080
                 - --leader-elect
+                env:
+                - name: RELATED_IMAGE_postgresql
+                  value: registry.redhat.io/rhel9/postgresql-15:latest
                 command:
                 - /manager
                 image: quay.io/rhdh/backstage-operator:v0.0.1
--- a/config/manager/default-config/db-statefulset.yaml
+++ b/config/manager/default-config/db-statefulset.yaml
@@ -28,8 +28,8 @@ spec:
               value: /var/lib/pgsql/data/userdata
           envFrom:
             - secretRef:
-                name: "{POSTGRESQL_SECRET}"  # will be replaced with 'backstage-psql-secrets-<cr-name>'  
-          image: quay.io/fedora/postgresql-15:latest
+                name: "{POSTGRESQL_SECRET}"  # will be replaced with 'backstage-psql-secrets-<cr-name>'
+          image: "${RELATED_IMAGE_postgresql}" # defined in operator-bundle CSV as an environment variable"
           imagePullPolicy: IfNotPresent
           securityContext:
             runAsNonRoot: true


config/manager/default-config/db-statefulset.yaml
```diff
@@ -28,8 +28,8 @@ spec:
               value: /var/lib/pgsql/data/userdata
           envFrom:
             - secretRef:
-                name: "{POSTGRESQL_SECRET}"  # will be replaced with 'backstage-psql-secrets-<cr-name>'  
-          image: quay.io/fedora/postgresql-15:latest
```
gazarenkov, line 32: I'd prefer to keep operator configuration in one place .
Why do we need it? (no, it is not hardcoded :) )

gazarenkov (changes requested):

config/manager/default-config/db-statefulset.yaml
nickboldt: yes, exactly. I move out the hardcoded value and replace it with the env var from the operator's CSV. This is what we implemented for Che Operator, it works great there.

gazarenkov: I mean I do not see config/manager/default-config/db-statefulset.yaml as a part of code...

Anyway, if cluster admin wants to change operator configuration (by any reason) it is possible to do that by configuring any part of default-backstage-configuration configMap generated from config/manager/default-config directory files. It contains all the objects taking part in Backstage runtime (yes, I owe documentation). 
I think this is quite popular approach and I'd consider it as more unified (it is not dependent on installation manner). 
We can consider bundle configuration as well but I am afraid several configuration points may be a bit confusing (and moving the whole files to CSV sounds even more scaring for me TBH).
In general, I do not reject your proposal, I'd just would like to weight cons and pros and know more about your perspective.

nickboldt: The point here is that if we're going to ship something that OOTB supports deploying a posgresql pod (like the helmchart does) then we need a way for this to work in airgapped environments. For airgap, we need:

* pinned digests (not tags) in image references, and
* all images that need to be present offline/airgapped must be declared in the CSV's `spec.relatedImages`. 

So... if we want to force configuring postgres in airgap to involve extra steps, well, cool. Maybe not ideal, but how doesn't love more manual steps? :D 

But if we want it to Just Work (following the standard OCP airgap instructions, eg., from https://access.redhat.com/documentation/en-us/red_hat_openshift_dev_spaces/3.10/html-single/administration_guide/index#installing-devspaces-in-a-restricted-environment ), then having the image reference be to an ENV var defined in the CSV and loaded by the operator into the container is another way to go.

gazarenkov: What I am trying to say is that (IMO) operator itself already has enough flexibility to :
a/ configure everything before deployment by changing manager/default-config/*.yaml files as needed and call 'make deploy' (or equivalent like 'kustomize ... | kubectl ...') note: there is nothing about changing the code/image, it is totally configuration task 
b/ changing configuration of anything "after deployment" changing configMap (like I mentioned above)
There re nothing special about postgres image, it is just a part of manifest.

But, if I understand you correct, it is not sufficient from productization standpoint because it requires configuration provided by CSV... Is it what you mean?

nickboldt: if you're going to preload a configmap with `image:` references then they need to be the SAME references as the container images in the CSV.

Otherwise you might end up with  quay.io/fedora/postgresql-15:latest =  quay.io/fedora/postgresql-15:1.0-123

but the CSV will have  quay.io/fedora/postgresql-15@sha256:somedigesthere which maps to  quay.io/fedora/postgresql-15:1.0-124

And when you try to launch a pod with `1.0-123` in it inside an airgapped cluster, you won't be able to pull that image... only `1.0-124`. You'll get the dreaded crash loop back off error, because the image you want to pull doesn't exist in the OCP container registry, and can't be found from the outside world due to the restricted environment.

So to avoid this problem, you need to have the SAME pinned image refs in your configmap as in your csv's `spec.relatedImages`.

That way an offline/airgapped/proxied cluster can be set up with all the images a user MIGHT NEED, inside the organization's firewall, and they can choose to enable/disable postgres via their configmap.

rm3l: @gazarenkov  Yes, that's my understanding to have out-of-the-box support for airgap installations: Since this is the [recommended way](https://docs.openshift.com/container-platform/4.14/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-restricted-network_osdk-generating-csvs) to leverage OCP's built-in airgap support, IMO we should rely on it, as it would alleviate the need for administrators to perform extra complicated steps in restricted envs.

But I think both approaches are not incompatible. As we discussed today, the goal would be to use, say, a placeholder everywhere we use images in these default YAML files.
So the ConfigMap would still be there, except that the operator code would need to be updated to look up these `RELATED_IMAGE_*` env vars (if set).
Note that we are already doing a similar thing with fields that we update at runtime when building the K8s objects to create (like when setting the actual resource names, or the postgres secret ref).
However, more advanced administrators can still change the ConfigMap at any time if they wish.

jianrongzhang89: @nickboldt @gazarenkov @rm3l I submitted a PR that implements this approach which was discussed during today's meeting (I renamed the env vars though):  https://github.com/janus-idp/operator/pull/86.

bundle/manifests/backstage-operator.clusterserviceversion.yaml
```diff
@@ -195,6 +195,9 @@ spec:
                 - --health-probe-bind-address=:8081
                 - --metrics-bind-address=127.0.0.1:8080
                 - --leader-elect
+                env:
```
jianrongzhang89, line 198: We need similar change for the backstage deployment object too.

config/manager/default-config/db-statefulset.yaml
gazarenkov: OK, 'make bundle' is smarter than me,  it generates default-config's configMap manifest (bundle/manifests/backstage-default-config_v1_configmap.yaml). 
So, admin just need to customize and apply this configuration, for whatever reason, not only airgap (for example to configure node toleration).
So, it seems, we already have what we need for that(except documentation) unless I missed something important?

BTW, to the point of not correct configuration (@nickboldt ), I think, we already have a good tool for admins to test modified config upfront. I mean CR.spec.rawConfig where one can just create own configmap with a whole or part of config and test before applying to operator.

@rm3l 
> Note that we are already doing a similar thing with fields that we update at runtime when building the K8s objects to create (like when setting the actual resource names, or the postgres secret ref).
However, more advanced administrators can still change the ConfigMap at any time if they wish.

I would not say it is the same thing, we've never substitute env variables and I'd really prefer not to build alternative template engine if we can avoid it. The guide you mentioned is correct for the case when image is hardcoded, which is not our case I believe.

gazarenkov (changes requested):

nickboldt:
closed in favour of #86

PR #85, opened 2023-12-19
Move `bundle.Dockerfile` under `docker` directory whenever it gets regenerated

### LINKED ISSUE #84: Put `bundle.Dockerfile` in `docker` directory when it gets regenerated
#75 moved `bundle.Dockerfile` under `docker`, but regenerating it with `make bundle` would currently recreate it in the root directory, creating confusion if it were to be pushed to the repo.
Worse, `make bundle-build` might not use the right `bundle.Dockerfile` after a regeneration.

This is a follow-up issue to https://github.com/janus-idp/operator/pull/79#discussion_r1431528458_

_Originally posted in https://github.com/janus-idp/operator/pull/79#discussion_r1431528458_

rm3l:
## Description
As reported in #84, `bundle.Dockerfile` gets recreated at the root directory whenever `make bundle` is run. This PR just moves the new file to the `docker` directory. Otherwise, it can create confusion if we have both `bundle.Dockerfile` files. Worse, `make bundle-build` might not use the right `bundle.Dockerfile` after a regeneration (since it uses `docker/bundle.Dockerfile` - https://github.com/janus-idp/operator/blob/main/Makefile#L258).

- Fixes #84 
- Related to #75 

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

`make bundle` should not create a ` bundle.Dockerfile` file at the root directory. It should exist under `docker/` instead.

### DIFF
--- a/docker/bundle.Dockerfile
+++ b/docker/bundle.Dockerfile
@@ -1,14 +1,3 @@
-# Copyright (c) 2023-2024 Red Hat, Inc.
-# This program and the accompanying materials are made
-# available under the terms of the Eclipse Public License 2.0
-# which is available at https://www.eclipse.org/legal/epl-2.0/
-#
-# SPDX-License-Identifier: EPL-2.0
-#
-# Contributors:
-#   Red Hat, Inc. - initial API and implementation
-#
-
 FROM scratch
 
 # Core bundle labels.
--- a/docker/bundle.Dockerfile
+++ b/docker/bundle.Dockerfile
@@ -1,3 +1,13 @@
+# Copyright (c) 2023-2024 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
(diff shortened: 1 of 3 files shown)


docker/bundle.Dockerfile
```diff
@@ -1,14 +1,3 @@
-# Copyright (c) 2023-2024 Red Hat, Inc.
```
rm3l, line 1: License header was after a regeneration. But I think it should be fine since we have a [LICENSE file](https://github.com/janus-idp/operator/blob/main/LICENSE) in the repo.

jianrongzhang89: @nickboldt is this fine?

nickboldt: well, the header is better. :D see my additional commit on this PR to make it reappear. 

Note also that `operator-sdk-v1.33.0` is out now, and throws some new warnings when run.

nickboldt (approved):

docker/bundle.Dockerfile
rm3l: Sounds good - thanks! We can look into fixing those warnings in a subsequent issue/PR.

PR #86, opened 2023-12-20
Enable Operator for restricted network environments

### LINKED ISSUE #80: Enable Operator for restricted network (air-gapped or proxied) environments

jianrongzhang89:
## Description
Use environment variables POSTGRESQL_IMAGE and BACKSTAGE_IMAGE to the operator container to indicate the default images used by the operator to deploy psql db and backstage application, respectively. 
Note that: 
a) If the psql db statefulset has image unset, or set to the special string {POSTGRESQL_IMAGE}, the operator will replace it with the value of the environment variable POSTGRESQL_IMAGE. Similarly BACKSTAGE_IMAGE is introdiced for the backstage deployment object.
b) The user can still override the backstage image using the Backstage CR spec.application.image field.

- Fixes #80 

## PR acceptance criteria

- [X] Tests
- [ ] Documentation

Deploy the operator and the examples CRs. Check that the backstage app and psql db pods are running successfully.

### DIFF
--- a/bundle/manifests/backstage-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/backstage-operator.clusterserviceversion.yaml
@@ -21,8 +21,8 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2023-12-18T21:01:24Z"
-    operators.operatorframework.io/builder: operator-sdk-v1.32.0
+    createdAt: "2023-12-20T02:05:49Z"
+    operators.operatorframework.io/builder: operator-sdk-v1.33.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
   name: backstage-operator.v0.0.1
   namespace: placeholder
@@ -164,6 +164,9 @@ spec:
                         operator: In
                         values:
                         - amd64
+                        - arm64
+                        - ppc64le
+                        - s390x
                       - key: kubernetes.io/os
                         operator: In
                         values:
@@ -198,6 +201,11 @@ spec:
                 - --leader-elect
                 command:
                 - /manager
+                env:
+                - name: POSTGRESQL_IMAGE
+                  value: quay.io/fedora/postgresql-15:latest
+                - name: BACKSTAGE_IMAGE
+                  value: quay.io/janus-idp/backstage-showcase:next
                 image: quay.io/rhdh/backstage-operator:v0.0.1
                 livenessProbe:
                   httpGet:
--- a/bundle/manifests/backstage-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/backstage-operator.clusterserviceversion.yaml
@@ -21,7 +21,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2023-12-20T02:05:49Z"
+    createdAt: "2023-12-20T13:27:15Z"
     operators.operatorframework.io/builder: operator-sdk-v1.33.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
   name: backstage-operator.v0.0.1
@@ -202,9 +202,9 @@ spec:
                 command:
                 - /manager
                 env:
-                - name: POSTGRESQL_IMAGE
+                - name: RELATED_IMAGE_postgresql
                   value: quay.io/fedora/postgresql-15:latest
-                - name: BACKSTAGE_IMAGE
+                - name: RELATED_IMAGE_backstage
                   value: quay.io/janus-idp/backstage-showcase:next
                 image: quay.io/rhdh/backstage-operator:v0.0.1
                 livenessProbe:
@@ -304,4 +304,9 @@ spec:
   provider:
     name: Red Hat Inc.
     url: https://www.redhat.com/
+  relatedImages:
+  - image: quay.io/fedora/postgresql-15:latest
+    name: postgresql
+  - image: quay.io/janus-idp/backstage-showcase:next
+    name: backstage
   version: 0.0.1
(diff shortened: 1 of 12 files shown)


bundle/manifests/backstage-operator.clusterserviceversion.yaml
```diff
@@ -198,6 +201,11 @@ spec:
                 - --leader-elect
                 command:
                 - /manager
+                env:
+                - name: POSTGRESQL_IMAGE
```
nickboldt, line 205: we might need to prefix these variables with `RELATED_IMAGE_` so that the downstream build can automatically collect them into `spec.relatedImages`.

controllers/backstage_controller_test.go
```diff
@@ -56,10 +56,12 @@ var _ = Describe("Backstage controller", func() {
...
-			Client:      k8sClient,
-			Scheme:      k8sClient.Scheme(),
-			Namespace:   ns,
-			OwnsRuntime: true,
+			Client:         k8sClient,
+			Scheme:         k8sClient.Scheme(),
+			Namespace:      ns,
+			OwnsRuntime:    true,
+			PsqlImage:      "quay.io/fedora/postgresql-15:latest",
+			BackstageImage: "quay.io/janus-idp/backstage-showcase:next",
```
nickboldt, line 64: if we use the restricted access (private repo) `quay.io/rhdh/rhdh=hub-rhel9:next` image here, we might be more easily able to test airgap, since OOTB a cluster won't be able to pull that image, but once copied inside the cluster's internal container registry, the image WILL be resolvable without the need for a quay pull secret.

nickboldt (approved):
Generally +1, but with one adjustment idea (which maybe makes more sense in the downstream version) and one question about the name of the env vars, which can be tested and fixed in a followup PR if needed.

bundle/manifests/backstage-operator.clusterserviceversion.yaml
jianrongzhang89: Sounds good. Updated the  PR.

controllers/backstage_controller_test.go
```diff
@@ -56,10 +56,12 @@ var _ = Describe("Backstage controller", func() {
...
 		backstageReconciler = &BackstageReconciler{
-			Client:      k8sClient,
-			Scheme:      k8sClient.Scheme(),
-			Namespace:   ns,
-			OwnsRuntime: true,
+			Client:         k8sClient,
+			Scheme:         k8sClient.Scheme(),
+			Namespace:      ns,
+			OwnsRuntime:    true,
+			PsqlImage:      "quay.io/fedora/postgresql-15:latest",
```
gazarenkov, line 63: Could you please specify how it was tested with restricted network environment?
As I understand understand (according to [previous discussion](https://github.com/janus-idp/operator/pull/83)) we specifically interested in testing SCV based installation.
Can you provide this information please.

gazarenkov (changes requested):

controllers/backstage_deployment.go
```diff
@@ -302,3 +304,16 @@ func (r *BackstageReconciler) validateAndUpdatePsqlSecretRef(backstage bs.Backst
 
 	return nil
 }
+
+func (r *BackstageReconciler) setDefaultDeploymentImage(deployment *appsv1.Deployment) {
+	for i, c := range deployment.Spec.Template.Spec.InitContainers {
```
rm3l, line 309: The same operation is done on several containers, coming from different parts of the Pod Template Spec.
There is a `visitContainers` function (accepting a `ContainerVisitor` function) that can be used for that specific purpose, I think. Maybe it can be used here instead?

controllers/local_db_statefulset.go
```diff
@@ -211,3 +215,11 @@ func (r *BackstageReconciler) patchLocalDbStatefulSetObj(statefulSet *appsv1.Sta
 
 	return nil
 }
+
+func (r *BackstageReconciler) setDefaultStatefulSetImage(statefulSet *appsv1.StatefulSet) {
+	for i, c := range statefulSet.Spec.Template.Spec.Containers {
```
rm3l, line 220: Same point for using the `visitContainers` function instead. And also, to cover the case where the StatefulSet Spec might contain Init Containers for example.

controllers/backstage_controller_test.go
nickboldt: airgap testing:
* start a cluster
* mirror images in the CSV to that cluster according to https://access.redhat.com/documentation/en-us/red_hat_openshift_dev_spaces/3.10/html-single/administration_guide/index#installing-devspaces-in-a-restricted-environment or https://docs.openshift.com/container-platform/4.13/installing/disconnected_install/installing-mirroring-installation-images.html
* "turn off the internet" via whatever proxy or DNS config works for you
* perform operator install 
* verify pods all start (including postgres and dev hub)
* verify RHDH loads (route works)

rm3l (approved):

controllers/local_db_statefulset.go
jianrongzhang89: Done

controllers/backstage_deployment.go
jianrongzhang89: Sounds good. Done

nickboldt:
Building in https://gitlab.cee.redhat.com/rhidp/rhdh/-/jobs/18503664

PR #94, opened 2023-12-21
Allow to disable automatic creation of Route on OpenShift

### LINKED ISSUE #93: Allow to disable automatic creation of Route on OpenShift
## User Story

As an OpenShift user, I want to disable the default behavior of the operator creating a Route by default for me, So that I can control how I want my Backstage instance to be exposed.
Since https://github.com/janus-idp/operator/pull/67, the operator can detect whether the current cluster is OpenShift or not. On OpenShift, it will automatically create a Route when reconciling a given CR. However some users may not want that behavior by default. So we should provide a way for them to skip the creation of this Route.

## Acceptance Criteria

- [ ] It should not change the current default behavior, i.e., it should create a Route if `spec.application.route` is unset.
- [ ] It should provide a way in the CR to completely disable the creation of a Route
  - [ ] For example, we can have a new `spec.application.route.enabled` boolean field (defaulting to `true`) in the CRD

rm3l:
## Description

This adds a new `spec.application.route.enabled` boolean field to the CRD. Its default value is `true` if not set.

- Fixes #93 

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

- This doesn't change the current default behavior, i.e.:
  -  if `spec.application` is unset or `spec.applicaiton.route` is not set, a Route is created:
```
$ oc apply -f examples/bs1.yaml
$ oc get route
NAME                     HOST/PORT                                                                          PATH   SERVICES                 PORT           TERMINATION     WILDCARD
backstage-bs1            backstage-bs1-my-ns.apps.rosa.pnyd6-mvq26-s6g.1ag4.p3.openshiftapps.com            /      backstage-bs1            http-backend   edge/Redirect   None
```

  - if `spec.application.route.enabled` is not set or is explicitly set to `true`, a Route is created:
```
$ oc apply -f examples/bs-route.yaml
$ oc get route
NAME                     HOST/PORT                                                                          PATH   SERVICES                 PORT           TERMINATION     WILDCARD
backstage-bs-route       test-bs-route.apps.rosa.pnyd6-mvq26-s6g.1ag4.p3.openshiftapps.com                  /      backstage-bs-route       http-backend   edge/Redirect   None
```

- A route is not created only if `spec.application.route.enabled` is `false`:
```
$ oc apply -f examples/bs-route-disabled.yaml
$ oc get route
$
```


controllers/backstage_route.go
```diff
@@ -110,3 +115,13 @@ func (r *BackstageReconciler) applyRouteParamsFromCR(route *openshift.Route, bac
 		}
 	}
 }
+
+func shouldCreateRoute(backstage bs.Backstage) bool {
+	if backstage.Spec.Application == nil {
```
jianrongzhang89, line 120: This can be simplified:
return backstage.Spec.Application == nil || backstage.Spec.Application.Route == nil || pointer.BoolDeref(backstage.Spec.Application.Route.Enabled, true)

jianrongzhang89 (approved):
Approved.

controllers/backstage_route.go
rm3l: Yes, I tried to favor readability, by not having a too long condition.

PR #95, opened 2023-12-21
Inject default Backend Auth Key app-config ConfigMap if `spec.application.appConfig.configMaps` is empty

rm3l:
## Description
This PR relaxes the condition for creating and injecting default app-config ConfigMap storing the default backend auth key for Backstage.
Previously, it would do so only if `spec.application.appConfig` and `spec.application.extraFiles` and `spec.application.extraEnvs` were not set.
Now, we are only checking that the list of user-defined app-configs in `spec.application.appConfig.configMaps` is non-empty.

&mdash;

## PR acceptance criteria

- [x] Tests
- [ ] Documentation

Previously, the following CR would not result in a running Backstage container:
```yaml
spec:
  application:
    appConfig:
      mountPath: /opt/app-root/src
    extraFiles:
      mountPath: /opt/app-root/src
    replicas: 1
  database:
    enableLocalDb: true
```

It should work now with the changes here.

### DIFF
--- a/controllers/backstage_backend_auth.go
+++ b/controllers/backstage_backend_auth.go
@@ -30,8 +30,7 @@ func (r *BackstageReconciler) getBackendAuthAppConfig(
 	backstage bs.Backstage,
 	ns string,
 ) (backendAuthAppConfig *bs.ObjectKeyRef, err error) {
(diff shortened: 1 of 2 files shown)


jianrongzhang89 (approved):
LGTM

PR #140, opened 2024-01-12
minKubeVersion should be set in the CSV

### LINKED ISSUE #139: minKubeVersion should be set in the CSV
minKubeVersion should be set in the CSV so that the operator does not get installed in old versions of OpenShift/Kubernetes.

jianrongzhang89:
## Description
Set minKuberVersion to 1.25.0 so that the operator can only be deployed on OpenShift 4.12+.

- Fixes #139 

## PR acceptance criteria

- [ X] Tests
- [ ] Documentation

Check the csv and confirm minKuberVersion is set.

### DIFF
--- a/bundle/manifests/backstage-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/backstage-operator.clusterserviceversion.yaml
@@ -21,7 +21,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2024-01-09T20:56:26Z"
+    createdAt: "2024-01-12T14:31:03Z"
     operators.operatorframework.io/builder: operator-sdk-v1.33.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
   name: backstage-operator.v0.0.1
@@ -301,6 +301,7 @@ spec:
   - email: jianrzha@redhat.com
     name: Jianrong Zhang
   maturity: alpha
+  minKubeVersion: 1.25.0
   provider:
     name: Red Hat Inc.
     url: https://www.redhat.com/
--- a/config/manifests/bases/backstage-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/backstage-operator.clusterserviceversion.yaml
@@ -47,6 +47,7 @@ spec:
   - email: jianrzha@redhat.com
     name: Jianrong Zhang
   maturity: alpha
+  minKubeVersion: 1.25.0
   provider:
     name: Red Hat Inc.
     url: https://www.redhat.com/


bundle/manifests/backstage-operator.clusterserviceversion.yaml
```diff
@@ -301,6 +301,7 @@ spec:
   - email: jianrzha@redhat.com
     name: Jianrong Zhang
   maturity: alpha
+  minKubeVersion: 1.26.0
```
rm3l, line 304: > Set minKuberVersion to 1.26.0 so that the operator can only be deployed on OpenShift 4.13+.

Why limit to 4.13+? I can see [here](https://access.redhat.com/support/policy/updates/openshift) that OCP 4.12 (and 4.11) are still in maintenance support, and the operator works fine on OCP 4.12 (I just tested it).

So I'd suggest setting it to 1.24.0 (which maps to OCP 4.11) instead (as done actually in the RHDH CSV: https://github.com/janus-idp/operator/blob/main/.rhdh/bundle/manifests/rhdh-operator.csv.yaml#L321), no?

```suggestion
  minKubeVersion: 1.24.0
```

config/manifests/bases/backstage-operator.clusterserviceversion.yaml
```diff
@@ -47,6 +47,7 @@ spec:
   - email: jianrzha@redhat.com
     name: Jianrong Zhang
   maturity: alpha
+  minKubeVersion: 1.26.0
```
rm3l, line 50: See https://github.com/janus-idp/operator/pull/140/files#r1450351386

bundle/manifests/backstage-operator.clusterserviceversion.yaml
rm3l: As discussed, it would make sense to set the minimum version to OCP 4.12

nickboldt, line 304:
```suggestion
  minKubeVersion: 1.25.0
```

config/manifests/bases/backstage-operator.clusterserviceversion.yaml
nickboldt:
```suggestion
  minKubeVersion: 1.25.0
```

nickboldt (approved):

jianrongzhang89:
The Quality Gate failure was not caused by this PR itself and this needs to be investigated separately.

rm3l (commented):
Should we change this line as well? https://github.com/janus-idp/operator/blob/main/.rhdh/bundle/manifests/rhdh-operator.csv.yaml#L321

nickboldt:
> Should we change this line as well? https://github.com/janus-idp/operator/blob/main/.rhdh/bundle/manifests/rhdh-operator.csv.yaml#L321

Done. https://github.com/janus-idp/operator/commit/a187eba092f1855cee8dd051718004b3658f8e94

Issue #72, opened 2023-12-14
Update existing resources when CR changes


rm3l:
- Estimation: 1w

Notes:
- Beware of secrets (should not change generated values)
- Beware of immutable fields

rm3l:
Notes from last meeting:
- before the first version of the Operator is released, we need to support the case where Backstage Config is changed (but not CR), as this is a common use case to update an `app-config` ConfigMap. We should document that the user needs to restart the Backstage pod in case of a config change (until #74 is implemented).

rm3l:
> @rm3l I probably missed that (being on PTO)
> 
> > we need to support the case where Backstage Config is changed (but not CR)
> 
> Not sure I understand terminology here but I suppose it means changes in app-config's (and extra) ConfigMaps etc, right?
> 

Yes, that's the (quite common) use case Tomas mentioned: making changes to app-config should be handled by the operator.

> I think we need carefully investigate the cases and understand and document limitations. For example, as I understand [kubernetes.io/docs/concepts/configuration/configmap/#mounted-configmaps-are-updated-automatically](https://kubernetes.io/docs/concepts/configuration/configmap/#mounted-configmaps-are-updated-automatically) it says it will be refreshed, but not for env variables and "A container using a ConfigMap as a [subPath](https://kubernetes.io/docs/concepts/storage/volumes#using-subpath) volume mount will not receive ConfigMap updates."
> 
> I would like we investigate, understand and document these limitations (what is supported by K8s and what we need to take care in the Operator) before we make a decision about making it supported user-friendly way.
> 
> Meanwhile, I agree, we need to mention that all the configuration changes "guaranteed" only after Backstage resource reconciliation (pod deleting or down/up scaling of replicas would be sufficient I guess)

Indeed, Tomas said that we are not sure Backstage would be able to hot-reload itself upon configuration changes; so we would need to restart Backstage even if K8s automatically refreshes the mounted files.

I agree we need to investigate. https://github.com/janus-idp/operator/issues/74 includes a research part to evaluate this. That's why we broke down the parent Epic (https://github.com/janus-idp/operator/issues/12) into two cases ([updates to the CR itself](https://github.com/janus-idp/operator/issues/72) vs [updates to resources linked to the CR](https://github.com/janus-idp/operator/issues/74)).
[...]

Issue #74, opened 2023-12-14
Support resource update when CR does not change but linked resources do


rm3l:
- Detect changes in app-config/extra-files ConfigMaps/Secrets
  - only watch for ConfigMaps/Secrets relevant for the operator
  - Requires some research + see impacts on system perf + load testing
- Trigger the reconciliation when detecting changes on such ConfigMaps or Secrets

## Links
- Parent Epic: https://github.com/janus-idp/operator/issues/12

rm3l:
Notes from last meeting:
- before the first version of the Operator is released, we need to support the case where Backstage Config is changed (but not CR), as this is a common use case to update an `app-config` ConfigMap. Until #74 is implemented, we could document that the user needs to restart the Backstage pod in case of a config change.

gazarenkov:
https://github.com/janus-idp/operator/issues/72#issuecomment-1884359945

gazarenkov:
fixed with https://github.com/janus-idp/operator/issues/236

PR #145, opened 2024-01-15
Update downstream RHDH docs now that CR updates are supported

rm3l:
## Description
This is a follow-up PR to https://github.com/janus-idp/operator/pull/119.

This removes the need to manually delete the Backstage Deployment when CR is updated.

- ref https://github.com/janus-idp/operator/issues/72

## PR acceptance criteria

- [ ] Tests
- [x] Documentation

See the doc changes here.


.rhdh/docs/openshift.adoc
```diff
@@ -26,6 +26,10 @@ image::images/rhdh_from_operator.png[RHDH from Operator]
 
 === Configurations for Operator-backed RHDH
 
+NOTE: At the moment, updates to the Backstage Custom Resource (CR) are automatically handled by the Operator.
```
gazarenkov, line 29: I do not think current implementation of update is correct, as discussed:
https://github.com/janus-idp/operator/pull/119#discussion_r1452064687

In general, an Operator must not deal with spec fields: 
"Types with both spec and status stanzas can (and usually should) have distinct authorization scopes for them. This allows users to be granted full write access to spec and read-only access to status, while relevant controllers are granted read-only access to spec but full write access to status."

And in this particular case, as stated https://kubernetes.io/docs/concepts/services-networking/cluster-ip-allocation/ if ClusterIP is dynamically assigned by contol plane (the case we are testing) we should not manage it during CR update. 

Thats the reason I asked [this PR](https://github.com/janus-idp/operator/pull/119) to be changed and so, we hardly can declare supporting Update cases.

gazarenkov (changes requested):

.rhdh/docs/openshift.adoc
```diff
@@ -26,6 +26,10 @@ image::images/rhdh_from_operator.png[RHDH from Operator]
...
+But updates to resources (e.g., ConfigMaps, Secrets) referenced in the CR are currently not handled automatically (unless the CR itself is updated).
+In this case, you will need to manually delete the Backstage Deployment, so that the Operator can recreate it with the updates.
```
nickboldt, line 31: delete the Deployment including pods, or just delete and recreate the CR? 

what about app config files ref'd by the CR?

.rhdh/docs/openshift.adoc
```diff
@@ -249,11 +248,7 @@ spec:
 ----
 
 . Click *Save*.
-. Navigate back to the *Topology* view and delete both the Backstage Deployment and PostgreSQL StatefulSet, so that the Operator can recreate it with the new changes to the CR.
-Note that this is a manual step at this time, but the operator will soon be able to handle such changes automatically.
-image:images/delete_backstage_deployment.png[Recreating RHDH Deployment]
-image:images/delete_postgres_statefulset.png[Deleting RHDH Postgresql Deployment]
-. Wait for the Red Hat Developer Hub Pod to start. The Database resource should not be created.
+. Navigate back to the *Topology* view and wait for the Red Hat Developer Hub Pod to start. The Database resource should not be created.
```
nickboldt, line 251:
```suggestion
. Navigate back to the *Topology* view and wait for the Red Hat Developer Hub Pod to start. The Postgres pod should not be created.
```

.rhdh/docs/openshift.adoc
rm3l: > delete the Deployment including pods, or just delete and recreate the CR?

Just deleting the Deployment, not the CR. As I've tested, the operator will recreate a Deployment from the CR.

rm3l: > what about app config files ref'd by the CR?

Yes, that was my point with this sentence: "But updates to resources (e.g., ConfigMaps, Secrets) referenced in the CR are currently not handled automatically (unless the CR itself is updated).
In this case, you will need to manually delete the Backstage Deployment, so that the Operator can recreate it with the updates."

So if only the app-config ConfigMap ref'd in the CR is updated, but not the CR itself, the user will have to manually delete the Backstage deployment (until https://github.com/janus-idp/operator/issues/74 is implemented).

nickboldt: got it -- wasn't sure if app config was a configmap object. 

You could also phrase this as "scaling down the deployment to 0 and then back to the previous number of pods", but I guess `manually delete the Backstage Deployment` is clear enough.

nickboldt (approved):

.rhdh/docs/openshift.adoc
rm3l: I remember I tried scaling down to 0 then up again, but that would depend on what changes are done in such linked resources. In some cases (e.g., adding a new app-config file as a new key to an existing ConfigMap), that would require updating the Deployment spec (e.g, main container args). So just scaling the existing Deployment wouldn't work.
I found deleting the Deployment so that the operator can recreate a new one fixed that kind of case.

PR #149, opened 2024-01-16
Add error messages to backstage CR status

### LINKED ISSUE #71: Add enough useful information in CR status
## User Story

As a user submitting a Custom Resource (CR), I want the operator to report back any relevant information to the CR status, So that I can have a better understanding of the status of my application by looking at the CR.
This can be useful in inspecting potential reconciliation issues or knowing if everything is fine from the operator's perspective.

## Acceptance Criteria

- [ ] When there are issues, it should return meaningful messages about the reason (like the status of the Backstage/PostgreSQL pods, e.g., ImagePullBackOff, CrashLoopBackOff, ...)
- [ ] It should report whether the application is Ready or not
- [ ] If Route is enabled in the CR, it should report the application route /URL once available
- [ ] ...

## Notes:
- Kubernetes API Conventions about `status`: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

jianrongzhang89:
## Description
When the operator fails to reconcile a CR, it should present an error in the CR's status for troubleshooting. 
When a CR has been successfully, the CR status should indicate that the deployment process is complete.

Simplified CR status conditions to use Deployed only.

Note that the running status support needs more investigation (on the possible impact on performance) and will be postponed to a future CR.

- Fixes #71 (partially)

## PR acceptance criteria

- [ X] Tests
- [ ] Documentation

Deploy example/bs1.yaml and the follow CR status should be displayed indicating sync completion.
```
status:
  conditions:
  - lastTransitionTime: "2024-01-23T15:06:21Z"
    message: ""
    reason: DeployOK
    status: "True"
    type: Deployed
``` 

Now update the CR, set application replicas to -1 and then apply the CR change:
```
apiVersion: janus-idp.io/v1alpha1
kind: Backstage
metadata:
  name: bs1
spec:
  application:
    replicas: -1
``` 
You should see the CR status like below:
```
status:
  conditions:
  - lastTransitionTime: "2024-01-23T15:08:24Z"
    message: 'failed to deploy Backstage Deployment: Deployment.apps "backstage-bs1"
      is invalid: spec.replicas: Invalid value: -1: must be greater than or equal
      to 0'
    reason: DeployFailed
    status: "False"
    type: Deployed
``` 

Now if you revert the CR to the default:
```
apiVersion: janus-idp.io/v1alpha1
kind: Backstage
metadata:
  name: bs1
spec:
``` 
You should see DeployOK again:
```
status:
  conditions:
  - lastTransitionTime: "2024-01-23T15:09:06Z"
    message: ""
    reason: DeployOK
    status: "True"
    type: Deployed
```

### DIFF
--- a/api/v1alpha1/backstage_types.go
+++ b/api/v1alpha1/backstage_types.go
@@ -20,14 +20,11 @@ import (
 
 const (
 	//RuntimeConditionRunning string = "RuntimeRunning"
-	ConditionSynced        string = "Synced"
-	ConditionRouteSynced   string = "RouteSynced"
-	ConditionLocalDbSynced string = "LocalDbSynced"
-	SyncOK                 string = "SyncOK"
-	SyncFailed             string = "SyncFailed"
-	Deleted                string = "Deleted"
-	EnvPostGresImage       string = "RELATED_IMAGE_postgresql"
-	EnvBackstageImage      string = "RELATED_IMAGE_backstage"
+	ConditionSynced   string = "Synced"
+	SyncOK            string = "SyncOK"
+	SyncFailed        string = "SyncFailed"
+	EnvPostGresImage  string = "RELATED_IMAGE_postgresql"
+	EnvBackstageImage string = "RELATED_IMAGE_backstage"
 )
 
 // BackstageSpec defines the desired state of Backstage
--- a/api/v1alpha1/backstage_types.go
+++ b/api/v1alpha1/backstage_types.go
@@ -19,15 +19,15 @@ import (
 )
 
 const (
-	RuntimeConditionRunning string = "RuntimeRunning"
-	RuntimeConditionSynced  string = "RuntimeSyncedWithConfig"
-	RouteSynced             string = "RouteSynced"
-	LocalDbSynced           string = "LocalDbSynced"
-	SyncOK                  string = "SyncOK"
-	SyncFailed              string = "SyncFailed"
-	Deleted                 string = "Deleted"
-	EnvPostGresImage        string = "RELATED_IMAGE_postgresql"
-	EnvBackstageImage       string = "RELATED_IMAGE_backstage"
+	//RuntimeConditionRunning string = "RuntimeRunning"
+	ConditionSynced        string = "Synced"
+	ConditionRouteSynced   string = "RouteSynced"
+	ConditionLocalDbSynced string = "LocalDbSynced"
+	SyncOK                 string = "SyncOK"
+	SyncFailed             string = "SyncFailed"
+	Deleted                string = "Deleted"
+	EnvPostGresImage       string = "RELATED_IMAGE_postgresql"
+	EnvBackstageImage      string = "RELATED_IMAGE_backstage"
 )
 
 // BackstageSpec defines the desired state of Backstage
--- a/api/v1alpha1/backstage_types.go
+++ b/api/v1alpha1/backstage_types.go
@@ -18,11 +18,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Constants for status conditions
 const (
-	//RuntimeConditionRunning string = "RuntimeRunning"
-	ConditionSynced   string = "Synced"
-	SyncOK            string = "SyncOK"
-	SyncFailed        string = "SyncFailed"
+	// TODO: RuntimeConditionRunning string = "RuntimeRunning"
+	ConditionSynced string = "Synced"
+	SyncOK          string = "SyncOK"
+	SyncFailed      string = "SyncFailed"
+)
+
+// Constants for image placeholders
+const (
 	EnvPostGresImage  string = "RELATED_IMAGE_postgresql"
 	EnvBackstageImage string = "RELATED_IMAGE_backstage"
 )
--- a/api/v1alpha1/backstage_types.go
+++ b/api/v1alpha1/backstage_types.go
@@ -21,10 +21,10 @@ import (
 // Constants for status conditions
 const (
 	// TODO: RuntimeConditionRunning string = "RuntimeRunning"
-	ConditionSynced string = "Synced"
-	SyncOK          string = "SyncOK"
-	SyncFailed      string = "SyncFailed"
-	SyncInProgress  string = "SyncInProgress"
+	ConditionDeployed string = "Deployed"
+	DeployOK          string = "DeployOK"
+	DeployFailed      string = "DeployFailed"
+	DeployInProgress  string = "DeployInProgress"
 )
 
 // Constants for image placeholders
--- a/api/v1alpha1/backstage_types.go
+++ b/api/v1alpha1/backstage_types.go
@@ -24,6 +24,7 @@ const (
 	ConditionSynced string = "Synced"
 	SyncOK          string = "SyncOK"
 	SyncFailed      string = "SyncFailed"
+	SyncInProgress  string = "SyncInProgress"
 )
 
 // Constants for image placeholders
--- a/controllers/backstage_controller_test.go
+++ b/controllers/backstage_controller_test.go
@@ -1445,6 +1445,39 @@ plugins: []
 				},
 			)
 		})
+
+		When("setting invalid replicas", func() {
+			var nbReplicas int32 = -1
+			var backstage *bsv1alpha1.Backstage
+			var crName = "backstage-invalid-replicas"
+
+			BeforeEach(func() {
+				backstage = buildBackstageCR(bsv1alpha1.BackstageSpec{
+					Application: &bsv1alpha1.Application{
+						Replicas: &nbReplicas,
+					},
+				})
+				backstage.Name = crName
+				err := k8sClient.Create(ctx, backstage)
+				Expect(err).To(Not(HaveOccurred()))
+			})
+
+			It("should fail to reconcile", func() {
+				By("Checking if the custom resource was successfully created")
+				Eventually(func() error {
+					found := &bsv1alpha1.Backstage{}
+					return k8sClient.Get(ctx, types.NamespacedName{Name: crName, Namespace: ns}, found)
+				}, time.Minute, time.Second).Should(Succeed())
+
+				By("Reconciling the custom resource created")
+				_, err := backstageReconciler.Reconcile(ctx, reconcile.Request{
+					NamespacedName: types.NamespacedName{Name: crName, Namespace: ns},
+				})
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).Should(ContainSubstring(
+					fmt.Sprintf("spec.replicas: Invalid value: %d: must be greater than or equal to 0", nbReplicas)))
+			})
+		})
 	})
 
 	Context("PostgreSQL", func() {
@@ -1526,14 +1559,14 @@ func findElementsByPredicate[T any](l []T, predicate func(t T) bool) (result []T
 }
 
 func isLocalDbDeployed(backstage bsv1alpha1.Backstage) bool {
-	if cond := meta.FindStatusCondition(backstage.Status.Conditions, bsv1alpha1.LocalDbSynced); cond != nil {
+	if cond := meta.FindStatusCondition(backstage.Status.Conditions, bsv1alpha1.ConditionLocalDbSynced); cond != nil {
 		return cond.Status == metav1.ConditionTrue && cond.Reason == bsv1alpha1.SyncOK
 	}
 	return false
 }
 
 func isSynced(backstage bsv1alpha1.Backstage) bool {
-	if cond := meta.FindStatusCondition(backstage.Status.Conditions, bsv1alpha1.RuntimeConditionSynced); cond != nil {
+	if cond := meta.FindStatusCondition(backstage.Status.Conditions, bsv1alpha1.ConditionSynced); cond != nil {
 		return cond.Status == metav1.ConditionTrue
 	}
 	return false
--- a/controllers/backstage_controller_test.go
+++ b/controllers/backstage_controller_test.go
@@ -371,14 +371,6 @@ var _ = Describe("Backstage controller", func() {
 			By("Checking the latest Status added to the Backstage instance")
 			verifyBackstageInstance(ctx)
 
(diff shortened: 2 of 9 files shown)


controllers/local_db_services.go
```diff
@@ -72,18 +72,20 @@ func (r *BackstageReconciler) reconcileLocalDbServices(ctx context.Context, back
...
 		},
 	}
-	if _, err := controllerutil.CreateOrUpdate(ctx, r.Client, service, r.serviceObjectMutFun(ctx, service, backstage, backstage.Spec.RawRuntimeConfig.LocalDbConfigName, configKey, serviceName, label)); err != nil {
+	if _, err := controllerutil.CreateOrUpdate(ctx, r.Client, service, r.serviceObjectMutFun(ctx, service, *backstage, backstage.Spec.RawRuntimeConfig.LocalDbConfigName, configKey, serviceName, label)); err != nil {
 		if errors.IsConflict(err) {
 			return fmt.Errorf("retry sync needed: %v", err)
 		}
-		return err
+		msg := fmt.Sprintf("failed to sync database service: %s", err.Error())
+		setStatusCondition(backstage, bs.ConditionSynced, metav1.ConditionFalse, bs.SyncFailed, msg)
```
gazarenkov, line 87: Why do we need 2 conditions making the same status and message on both?

controllers/backstage_route.go
```diff
@@ -49,10 +49,12 @@ func (r *BackstageReconciler) reconcileBackstageRoute(ctx context.Context, backs
...
 		}
-		setStatusCondition(backstage, bs.RouteSynced, metav1.ConditionFalse, bs.SyncFailed, fmt.Sprintf("Error:%s", err.Error()))
-		return err
+		msg := fmt.Sprintf("failed to sync Backstage Route: %s", err.Error())
+		setStatusCondition(backstage, bs.ConditionSynced, metav1.ConditionFalse, bs.SyncFailed, msg)
+		setStatusCondition(backstage, bs.ConditionRouteSynced, metav1.ConditionFalse, bs.SyncFailed, msg)
+		return fmt.Errorf(msg)
 	}
-	setStatusCondition(backstage, bs.RouteSynced, metav1.ConditionTrue, bs.SyncOK, fmt.Sprintf("Route host:%s", route.Spec.Host))
+	setStatusCondition(backstage, bs.ConditionRouteSynced, metav1.ConditionTrue, bs.SyncOK, fmt.Sprintf("Route host: %s", route.Spec.Host))
```
gazarenkov, line 57: Can not we have single "running" condition for backstage and contribute there the state and the message?
Indeed, if Backstage pod is running well but you can not route it, does it mean Backstage is working as expected?

api/v1alpha1/backstage_types.go
```diff
@@ -19,15 +19,15 @@ import (
...
-	RuntimeConditionSynced  string = "RuntimeSyncedWithConfig"
-	RouteSynced             string = "RouteSynced"
-	LocalDbSynced           string = "LocalDbSynced"
-	SyncOK                  string = "SyncOK"
-	SyncFailed              string = "SyncFailed"
-	Deleted                 string = "Deleted"
-	EnvPostGresImage        string = "RELATED_IMAGE_postgresql"
-	EnvBackstageImage       string = "RELATED_IMAGE_backstage"
+	//RuntimeConditionRunning string = "RuntimeRunning"
+	ConditionSynced        string = "Synced"
```
gazarenkov, line 23: "Synced" is confusing in all of those 2 IMO, in fact, it states that runtime objects are deployed/working/running... 

Note: Initially "Synced" indicated that runtime is in sync with CR (CR "owned" runtime).

gazarenkov (changes requested):

api/v1alpha1/backstage_types.go
```diff
@@ -19,15 +19,15 @@ import (
...
-	RuntimeConditionRunning string = "RuntimeRunning"
-	RuntimeConditionSynced  string = "RuntimeSyncedWithConfig"
-	RouteSynced             string = "RouteSynced"
-	LocalDbSynced           string = "LocalDbSynced"
-	SyncOK                  string = "SyncOK"
-	SyncFailed              string = "SyncFailed"
-	Deleted                 string = "Deleted"
-	EnvPostGresImage        string = "RELATED_IMAGE_postgresql"
-	EnvBackstageImage       string = "RELATED_IMAGE_backstage"
+	//RuntimeConditionRunning string = "RuntimeRunning"
```
gazarenkov, line 22: IMO RuntimeConditionRunning is exactly what you are doing with XSynced conditions. It is about "Running" (Deployed|Working), not about "Synced" :)

api/v1alpha1/backstage_types.go
```diff
@@ -19,15 +19,15 @@ import (
...
+	ConditionSynced        string = "Synced"
+	ConditionRouteSynced   string = "RouteSynced"
+	ConditionLocalDbSynced string = "LocalDbSynced"
+	SyncOK                 string = "SyncOK"
+	SyncFailed             string = "SyncFailed"
+	Deleted                string = "Deleted"
+	EnvPostGresImage       string = "RELATED_IMAGE_postgresql"
```
gazarenkov, line 29: Those 2 are not part of the API.

controllers/backstage_controller_test.go
```diff
@@ -1445,6 +1445,39 @@ plugins: []
 				},
 			)
 		})
+
+		When("setting invalid replicas", func() {
```
gazarenkov, line 1449: What are we testing by this 30 lines of code? 
this code:
deployment.Spec.Replicas = backstage.Spec.Application.Replicas
or
Kubernete's ability to process a negative amount of replicas? :)

gazarenkov (changes requested):
In general, existence of basic comments / descriptions would significantly simplify understanding
As mentioned in comments:
- Making multiple status conditions is discussible (better start simple)
- Name of "..Synced" is confusing for proposed conditions' functions
- The purpose of the test added is not that clear

api/v1alpha1/backstage_types.go
jianrongzhang89: > "Synced" is confusing in all of those 3 IMO, in fact, it states that runtime objects are deployed/working/running...
> 
> Note: Initially "Synced" indicated that runtime is in sync with CR (CR "owned" runtime).
I borrowed the concept "Synced" from the Argo CD Operator which has been widely used in the community.
3 status conditions are used because it allows users whether the localdb has been successfully "synced" and whether the route has been created along with the route host information. It is hard to use a single condition type as the later status overrides the previous one.
In addition, it is pretty common to use separate condition types, for instance, in the k8s deployment object, we can see that both "Progressing" which indicates the deployment status and "Available" that is to indicate the runtime status are used.

controllers/backstage_route.go
jianrongzhang89: The sync status and runtime object status must be handled separately. If we only use one condition, it would be complicated to handle the situation such as: 1) the backstage CR is deployed/synced and and backstage app is up and running, and then 2) the CR is updated with an incorrect configuration and  sync would fail, but the backstage application is still up and running (since the CR change is not applied).

controllers/backstage_controller_test.go
jianrongzhang89: The purpose is that when an invalid configuration is applied an error message should be logged in the CR status.

controllers/local_db_services.go
jianrongzhang89: You are right, the error messages are duplicated. The two conditions are used as one is to indicate whether localdb has been successfully deployed while the other is indicate whether all resources have been deployed.

api/v1alpha1/backstage_types.go
jianrongzhang89: As we discussed in #71 , the runtime object status will be deferred to a later release.

api/v1alpha1/backstage_types.go
gazarenkov: "Sync" does make sense in GitOps (ArgoCD) case where you have desired state remotely (git repository) which is not our case.

I am not debating number of conditionals in general, I mean is seems to me it is too much for our case, I'd prefer to start smaller until necessity of more is proven

controllers/backstage_route.go
gazarenkov: yes, it is two different statuses (we've had it) but in my definitions:
- synced means CR state is synced with runtime 
- running means runtime is functional

jianrongzhang89: > yes, it is two different statuses (we've had it) but in my definitions:
> 
> * synced means CR state is synced with runtime
> * running means runtime is functional
Yeah this sounds good to me.  As you suggested, we can wait for feedback from customers and see how  "Running" status should be implemented.

api/v1alpha1/backstage_types.go
gazarenkov: > As we discussed in #71 , the runtime object status will be deferred to a later release.

Well, what I can see in this regard is your recent comment proposing to defer it, which does not look like "we discussed" :) .
Let's discuss, sure!

jianrongzhang89: @gazarenkov Thanks!

controllers/local_db_services.go
gazarenkov: Yes, I see that, and my suggestion is to have only one with informative enough message.

controllers/backstage_controller_test.go
gazarenkov: Well, I can see:
When("setting invalid replicas"
It("Should fail to reconsile...

BTW, if it is just to check the status string could you please consider checking it in some else testcase instead (it is already too overloaded as you can see).

gazarenkov (changes requested):

controllers/backstage_controller_test.go
jianrongzhang89: Do you mean to move it into part of another test case? Thanks.

controllers/local_db_services.go
jianrongzhang89: This sounds good to me. I removed the RouteSynced and LocalDBSynced conditions from the status. We can add further enhancements in the future if needed.

gazarenkov (changes requested):
I've added one, more detailed proposals to this [comment](https://github.com/janus-idp/operator/issues/71#issuecomment-1899868209).
@jianrongzhang89 please consider it, hope it better explains my concerns

api/v1alpha1/backstage_types.go
```diff
@@ -19,15 +19,12 @@ import (
...
-	SyncOK                  string = "SyncOK"
-	SyncFailed              string = "SyncFailed"
-	Deleted                 string = "Deleted"
-	EnvPostGresImage        string = "RELATED_IMAGE_postgresql"
-	EnvBackstageImage       string = "RELATED_IMAGE_backstage"
+	//RuntimeConditionRunning string = "RuntimeRunning"
+	ConditionSynced   string = "Synced"
+	SyncOK            string = "SyncOK"
+	SyncFailed        string = "SyncFailed"
+	EnvPostGresImage  string = "RELATED_IMAGE_postgresql"
```
rm3l, line 26: nit: can we move the `Env*Image` constants into a separate `const` group? I find it more readable to group related declarations..

controllers/local_db_statefulset.go
```diff
@@ -262,7 +263,7 @@ func (r *BackstageReconciler) cleanupLocalDbResources(ctx context.Context, backs
 		},
 	}
 	if _, err := r.cleanupResource(ctx, secret, backstage); err != nil {
-		return err
+		return fmt.Errorf("failed to delete generated database secret, reason: %s", err)
```
rm3l, line 266: General suggestion (but can be handled consistently in a separate PR): could we wrap original errors instead, so that we don't lose the underlying errors?

```suggestion
		return fmt.Errorf("failed to delete generated database secret, reason: %w", err)
```

api/v1alpha1/backstage_types.go
jianrongzhang89: I simplified the solution. Please take a look again.

api/v1alpha1/backstage_types.go
jianrongzhang89: Good idea. Done.

jianrongzhang89:
> I've added one, more detailed proposals to this [comment](https://github.com/janus-idp/operator/issues/71#issuecomment-1899868209). @jianrongzhang89 please consider it, hope it better explains my concerns

@gazarenkov I added some comments to https://github.com/janus-idp/operator/issues/71. I believe the current implementation is in align with your proposal. Let me know if there are any more issues.

controllers/local_db_statefulset.go
jianrongzhang89: Lets handle this as a separate PR as additional thoughts are needed because the information we store in the status condition should be concise (ie, putting the full call stack is not a good idea), and on the other hand, we want the user to have as much information as possible.

controllers/backstage_controller_test.go
gazarenkov: Not exactly, I rather mean if the purpose is to check if the status is set correctly, whatever the reason of failing is, you probably can reuse some other "should fail to reconcile..." case instead...

Well, my point is that testing negative replicas number (along with status printing) you IMO rather test k8s than backstage operator.
Maybe it makes sense to find something other to test or I missed something?

jianrongzhang89: The intent of this test was not to test k8s itself, and it is used to verify that an error can appear in the status, regardless it is caused by the operator itself or caused by the underlying k8s. So I think it should be fine here.

jianrongzhang89:
@gazarenkov @rm3l @nickboldt I changed the CR status type to "Deployed". Could you please take a look again? Thanks.

rm3l (approved):
Sounds good to me, after testing.

/hold

Approved but put on hold to prevent auto-merge.
@gazarenkov feel free to remove the hold if it's fine with you.

rm3l:
/hold cancel

Removing the hold as discussed, to get this merged.

Issue #151, opened 2024-01-17
Enable gosec scanning


rm3l:
- Add github action to workflow
- Handle findings

rm3l:
/close

Fixed by https://github.com/janus-idp/operator/pull/155

PR #152, opened 2024-01-17
Do not fallback to a default hard-coded password for PostgreSQL if generation fails

rm3l:
## Description

This has been flagged as a potential security issue, which we should avoid doing.

https://github.com/janus-idp/operator/issues/151 will make sure to enforce such checks and detect such issues earlier as we move forward.

cc @kim-tsao 

&mdash;

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

`kubectl apply -f examples/bs1.yaml` should result in a working Backstage instance.

### DIFF
--- a/controllers/local_db_secret.go
+++ b/controllers/local_db_secret.go
@@ -16,13 +16,14 @@ package controller
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/base64"
 	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
-	k8srand "k8s.io/apimachinery/pkg/util/rand"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	bs "janus-idp.io/backstage-operator/api/v1alpha1"
@@ -54,7 +55,10 @@ func (r *BackstageReconciler) handlePsqlSecret(ctx context.Context, statefulSet 
 			return nil, fmt.Errorf("failed to get secret for PostgreSQL DB (%q), reason: %s", secretName, err)
 		}
 		// Create a secret with a random value
-		pwd := k8srand.String(24)
+		pwd, pwdErr := generatePassword(24)
+		if pwdErr != nil {
+			return nil, fmt.Errorf("failed to generate a password for the PostgreSQL database: %w", pwdErr)
+		}
 		sec.StringData["POSTGRES_PASSWORD"] = pwd
 		sec.StringData["POSTGRESQL_ADMIN_PASSWORD"] = pwd
 		sec.StringData["POSTGRES_HOST"] = getDefaultDbObjName(*backstage)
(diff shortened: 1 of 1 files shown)


gazarenkov (commented):
Would be great to have a link/quote to CVE or other regulation for reference.

controllers/local_db_secret.go
```diff
@@ -57,16 +55,12 @@ func (r *BackstageReconciler) handlePsqlSecret(ctx context.Context, statefulSet
...
-			bytes := make([]byte, length)
-			if _, randErr := rand.Read(bytes); randErr != nil {
-				// Do not fail, but use a fallback value
-				return _defaultPostGresSecretValue
-			}
-			return base64.StdEncoding.EncodeToString(bytes) // Encode the password to prevent special characters
-		}(24)
-		sec.StringData["POSTGRES_PASSWORD"] = val
-		sec.StringData["POSTGRESQL_ADMIN_PASSWORD"] = val
+		pwd, pwdErr := generatePassword(24)
```
gazarenkov, line 58: I guess there are no needs in a function throwing error,
e g consider using "k8s.io/apimachinery/pkg/util/rand" 
like: rand.String(24)

kim-tsao:
> Would be great to have a link/quote to CVE or other regulation for reference.

@gazarenkov, this was discovered by gosec scanning (which I plan to enable on this repo after this PR is merged). See https://securego.io/docs/rules/g101.  While using defaults is not prohibited, we need to make sure end users are aware the defaults are assigned and we need to document that they need to change passwords immediately.  In this case, the risk is that users may not take action, especially if we are handling the error and silently assigning a value for them.

controllers/local_db_secret.go
rm3l: That makes sense to use it. I didn't know this function - thanks!

rm3l: Done in [`bbab5ab` (#152)](https://github.com/janus-idp/operator/pull/152/commits/bbab5ab1c94b85202e57e40b3d570969e36d64c4)

kim-tsao: can `k8s.io/apimachinery/pkg/util/rand` be used for cryptographic purposes?  FYI `crypto/rand` is recommended here: https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html

kim-tsao: looks like it's not...k8s uses math rand which is unsafe

rm3l: > looks like it's not...k8s uses math rand which is unsafe

So you're suggesting to revert to `crypto/rand`?

controllers/local_db_secret.go
```diff
@@ -1,37 +1,34 @@
...
 	"context"
-	"crypto/rand"
-	"encoding/base64"
 	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	k8srand "k8s.io/apimachinery/pkg/util/rand"
```
kim-tsao, line 25: please change back to `crypto/rand` as this is unsafe to use

kim-tsao (changes requested):

controllers/local_db_secret.go
rm3l: Okay - reverted to the previous implementation using `crypto/rand` in [`abf657e` (#152)](https://github.com/janus-idp/operator/pull/152/commits/abf657e49bc153c1649bf2f648657cac856231c3)

controllers/local_db_secret.go
rm3l: Done in [`abf657e` (#152)](https://github.com/janus-idp/operator/pull/152/commits/abf657e49bc153c1649bf2f648657cac856231c3)

gazarenkov (commented):
https://github.com/janus-idp/operator/pull/152#discussion_r1457238844

gazarenkov (approved):
OK with me, I just would add more detailed message b/c this secret is very internal

rm3l:
> OK with me, I just would add more detailed message b/c this secret is very internal

Where would you add a more detailed message? This secret is used as the DB password, so it might not be that internal in case the DB Service is exposed (accidentally or not).

PR #155, opened 2024-01-19
Enable gosec scanning

kim-tsao:
## Description

- Adds the scanner to our github workflow so it can run as part of our PR checks
- Handle initial findings

- Fixes #_issue_number_
-
https://github.com/janus-idp/operator/issues/151

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -24,6 +24,22 @@ jobs:
     - name: Test
       run: make test
 
+    - name: Run Gosec Security Scanner
+      run: |
+        go install github.com/securego/gosec/v2/cmd/gosec@v2.18.2
+        make gosec
+        if [[ $? != 0 ]]
+        then
+          echo "gosec scanner failed to run "
+          exit 1
+        fi
+
+    - name: Upload SARIF file
+      uses: github/codeql-action/upload-sarif@v2
+      with:
+        # Path to SARIF file relative to the root of the repository
+        sarif_file: gosec.sarif
+
   container-image-build:
     runs-on: ubuntu-latest
 
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -25,14 +25,7 @@ jobs:
       run: make test
 
     - name: Run Gosec Security Scanner
-      run: |
-        go install github.com/securego/gosec/v2/cmd/gosec@v2.18.2
-        make gosec
-        if [[ $? != 0 ]]
-        then
-          echo "gosec scanner failed to run "
-          exit 1
-        fi
+      run: make gosec
 
     - name: Upload SARIF file
       uses: github/codeql-action/upload-sarif@v2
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -111,6 +111,14 @@ fmt_license: addlicense ## Ensure the license header is set on all files
 	$(ADDLICENSE) -v -f license_header.txt $$(find . -not -path '*/\.*' -name '*.go')
 	$(ADDLICENSE) -v -f license_header.txt $$(find . -name '*ockerfile')
 
+.PHONY: gosec
+gosec: ## run the gosec scanner for non-test files in this repo
+  	ifeq ($(shell command -v gosec 2> /dev/null),)
```
rm3l, line 116: To make sure we all use the same version (and do not use any `gosec` installed globally on the system), could you download it locally (to `$(LOCALBIN)`) if necessary? See how we are handling other tool binaries like [`golangci-lint`](https://github.com/janus-idp/operator/blob/main/Makefile#L235-L238) or [`controller-gen`](https://github.com/janus-idp/operator/blob/81e7506b3c86fb7b6a374a28649e27df6ba5a9fb/Makefile#L225-L228).

Makefile
```diff
@@ -111,6 +111,14 @@ fmt_license: addlicense ## Ensure the license header is set on all files
...
+	  	$(error "gosec must be installed for this rule: go install github.com/securego/gosec/v2/cmd/gosec@v2.18.2" && exit 1)
+  	endif
+  	# we let the report content trigger a failure using the GitHub Security features.
+	gosec -no-fail -fmt sarif -out gosec.sarif  ./...
```
rm3l, line 120: Can the output format be an option, so that we can override it when calling `make gosec`?

rm3l: Also, please add `gosec.sarif` to the `.gitignore` file..

.github/workflows/pr.yaml
```diff
@@ -24,6 +24,22 @@ jobs:
     - name: Test
       run: make test
 
+    - name: Run Gosec Security Scanner
+      run: |
+        go install github.com/securego/gosec/v2/cmd/gosec@v2.18.2
+        make gosec
+        if [[ $? != 0 ]]
```
rm3l, line 31: Do we need this check? If the exit code of `make gosec` is not 0, the action should fail, no?

rm3l (changes requested):

Makefile
kim-tsao: FYI, I locked it down to a specific version because pulling the latest has broken us in the past especially in areas with go version compatibility.  We should update regularly when we can (and rely on renovate to help us here once that's enabled)

.github/workflows/pr.yaml
kim-tsao: yes, it's redundant since I added the check in the make rule.  I'll remove it

Makefile
kim-tsao: I totally agree, this is a better approach

gazarenkov (approved):
LGTM

PR #181, opened 2024-01-31
chore: label every new issue with jira label

kadel:
to make sure that every new issue is recorded in Jira as well


rm3l (approved):

nickboldt:
counter argument / unpopular opinion since this is moving downstream into RH developer org:

Why not disable GH issues for this repo, and make people use JIRA directly?

PR #183, opened 2024-02-01
feat: new script for restricted env setup -...

nickboldt:
### What does this PR do?

feat: new script for restricted env setup - fetch dev hub images and related images from the index, and mirror to a cluster's internal registry
TODO: fix the skopeo copy step - not working :(

Signed-off-by: Nick Boldt <nboldt@redhat.com>

### Screenshot/screencast of this PR
N/A

### What issues does this PR fix or reference?
N/A (or see commit message above for issue number)

### How to test this PR?
N/A

### PR Checklist

[As the author of this Pull Request I made sure that:](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#pull-request-template-and-its-checklist)

- [x] [The Eclipse Contributor Agreement is valid](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-eclipse-contributor-agreement-is-valid)
- [x] [Code produced is complete](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-produced-is-complete)
- [x] [Code builds without errors](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-builds-without-errors)
- [ ] [Tests are covering the bugfix](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#tests-are-covering-the-bugfix)
- [x] [The repository devfile is up to date and works, if one exists](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-repository-devfile-is-up-to-date-and-works)
- [x] [Sections What issues does this PR fix or reference and How to test this PR completed](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#sections-what-issues-does-this-pr-fix-or-reference-and-how-to-test-this-pr-completed)
- [x] [Relevant user documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [x] [Relevant contributing documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [x] [CI/CD changes implemented, documented and communicated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#cicd-changes-implemented-documented-and-communicated)

### Reviewers

Reviewers, please comment how you tested the PR when approving it.

### DIFF
--- a/scripts/prepare-restricted-environment.sh
+++ b/scripts/prepare-restricted-environment.sh
@@ -1,0 +1,132 @@
+#!/bin/bash
+#
+# Copyright (c) 2024 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+# Fail on error
+set -e
+
+# INTERNAL_REGISTRY_URL, eg 'ec2-3-12-71-143.us-east-2.compute.amazonaws.com:5000' or 'default-route-openshift-image-registry.apps.ci-ln-x0yk982-72292.origin-ci-int-gce.dev.rhcloud.com'
+# INTERNAL_REG_USERNAME, eg., dummy
+# INTERNAL_REG_PASSWORD, eg., dummy
+
+# podman login registry.redhat.io -u ${RRIO_USERNAME} -p ${RRIO_PASSWORD}
+# podman login ${INTERNAL_REGISTRY_URL} -u ${INTERNAL_REG_USERNAME} -p ${INTERNAL_REG_PASSWORD} --tls-verify=false
+
+# example usage:
+# ./prepare-restricted-environment.sh \
+#   --prod_operator_index "registry.redhat.io/redhat/redhat-operator-index:v4.14" \
+#   --prod_operator_package_name "devspaces" \
+#   --prod_operator_bundle_name "rhdh-operator" \
+#   --prod_operator_version "v1.1.0" \
+#   --my_registry "$INTERNAL_REGISTRY_URL"
+while [ $# -gt 0 ]; do
+  if [[ $1 == *"--"* ]]; then
+    param="${1/--/}"
+    declare "$param"="$2"
+  fi
+  shift
+done
+
+# Display commands
+# set -x
+
+# Operators
+declare prod_operator_index="${prod_operator_index:?Must set --prod_operator_index: for OCP 4.12, use registry.redhat.io/redhat/redhat-operator-index:v4.12 or quay.io/rhdh/iib:latest-v4.14-x86_64}"
+declare prod_operator_package_name="rhdh"
+declare prod_operator_bundle_name="rhdh-operator"
+declare prod_operator_version="${prod_operator_version:?Must set --prod_operator_version: for stable channel, use v1.1.0; for stable-1.1 channel, use v1.1.1}" # eg., v1.1.0 or v1.1.1
+
+# Destination registry
+declare my_registry="${my_registry:?Must set --my_registry: something like 'default-route-openshift-image-registry.apps.ci-ln-x0yk982-72292.origin-ci-int-gce.dev.rhcloud.com' for your cluster}"
+declare my_operator_index_image_name_and_tag=${prod_operator_package_name}-index:${prod_operator_version}
+declare my_operator_index="${my_registry}/${prod_operator_package_name}/${my_operator_index_image_name_and_tag}"
+
+declare my_catalog=${prod_operator_package_name}-disconnected-install
+declare k8s_resource_name=${my_catalog}
+
+# from https://docs.openshift.com/container-platform/4.14/registry/securing-exposing-registry.html
+echo "[INFO] Expose the default registry and log in ..."
+oc patch configs.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
+HOST=$(oc get route default-route -n openshift-image-registry --template='{{ .spec.host }}')
+echo "[INFO] Default registry is $HOST"
+my_registry="$HOST"
+
+# if able to run sudo and update your ca trust store
+# oc get secret -n openshift-ingress  router-certs-default -o go-template='{{index .data "tls.crt"}}' | base64 -d | sudo tee /etc/pki/ca-trust/source/anchors/${HOST}.crt  > /dev/null
+# sudo update-ca-trust enable
+# sudo podman login -u kubeadmin -p "$(oc whoami -t)" "$HOST"
+
+# else just login with tls disabled
+podman login -u kubeadmin -p "$(oc whoami -t)" --tls-verify=false "$HOST"
+
+# Create local directory
+mkdir -p "${my_catalog}/${prod_operator_package_name}"
+
+echo "[INFO] Fetching metadata for the ${prod_operator_package_name} operator catalog channel, packages, and bundles."
+opm render "${prod_operator_index}" \
+  | jq "select \
+    (\
+      (.schema == \"olm.bundle\" and .name == \"${prod_operator_bundle_name}.${prod_operator_version}\") or \
+      (.schema == \"olm.package\" and .name == \"${prod_operator_package_name}\") or \
+      (.schema == \"olm.channel\" and .package == \"${prod_operator_package_name}\") \
+    )" \
+  | jq "select \
+     (.schema == \"olm.channel\" and .package == \"${prod_operator_package_name}\").entries \
+      |= [{name: \"${prod_operator_bundle_name}.${prod_operator_version}\"}]" \
+  > "${my_catalog}/${prod_operator_package_name}/render.json"
+
+echo "[DEBUG] Got $(cat "${my_catalog}/${prod_operator_package_name}/render.json" | wc -l) lines of JSON from the index!"
+# echo "[DEBUG] Got this from the index:
+# ======"
+# cat "${my_catalog}/${prod_operator_package_name}/render.json"
+# echo "======"
+
+echo "[INFO] Creating the catalog dockerfile."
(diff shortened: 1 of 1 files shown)


scripts/prepare-restricted-environment.sh
```diff
@@ -0,0 +1,132 @@
...
+echo "[INFO] Disabling the default Red Hat Ecosystem Catalog."
+oc patch OperatorHub cluster --type json \
+    --patch '[{"op": "add", "path": "/spec/disableAllDefaultSources", "value": true}]'
+
+echo "[INFO] Deploying your catalog image to the $my_operator_index registry."
+# See: https://docs.openshift.com/container-platform/latest/installing/disconnected_install/installing-mirroring-installation-images.html#olm-mirroring-catalog_installing-mirroring-installation-images
+### TODO fix this step or switch to oc adm catalog mirror?
+# https://docs.openshift.com/container-platform/4.14/installing/disconnected_install/installing-mirroring-installation-images.html#installation-images-samples-disconnected-mirroring-assist_installing-mirroring-installation-images ?
+# FATA[0010] writing manifest: uploading manifest v1.1.0 to default-route-openshift-image-registry.apps.ci-ln-x0yk982-72292.origin-ci-int-gce.dev.rhcloud.com/rhdh/rhdh-index: denied 
+# skopeo copy --src-tls-verify=false --dest-tls-verify=false --all "containers-storage:$my_operator_index" "docker://$my_operator_index"
```
rm3l, line 108: I got the same "denied" error, and after looking into the OCP Image Registry logs, I found that it was expecting a namespace `rhdh` to be available, in order to be able to push an image named `default-route-openshift-image-registry.apps.ci-ln-x0yk982-72292.origin-ci-int-gce.dev.rhcloud.com/rhdh/rhdh-index`.
I had to do the same thing for all the images for we are calling skopeo:
```
for ns in rhdh rhel9 openshift4 rh-osbs; do oc create namespace $ns; done
```

But even after that, it made some progress, but then I hit another issue:
```
FATA[0015] copying image 1/4 from manifest list: writing signatures: downloading signatures for sha256:5ef1720092d3b8431c7eee5a5ffb336a9a3d6329cc17ee4aaa1a2f18b42267ef in default-route-opensh
ift-image-registry.apps.7737beef7661f70d0e22.hypershift.aws-2.ci.openshift.org/rhel9/postgresql-15: StatusCode: 404, ""
```
Then I tried to work around this error by adding `--remove-signatures` to skopeo, but I hit another error for some of the images:
```
"application/vnd.oci.image.index.v1+json" to be written to destination, but we cannot modify it: "Instructed to preserve digests"
```

So to me, it seems we cannot use the internal OpenShift image registry as a target registry, as mentioned in the docs: https://docs.openshift.com/container-platform/4.14/installing/disconnected_install/installing-mirroring-installation-images.html#installation-about-mirror-registry_installing-mirroring-installation-images

---

After taking a closer look at the [`prepare-restricted-environment.sh` script from DevSpaces
](https://access.redhat.com/webassets/avalon/d/Red_Hat_OpenShift_Dev_Spaces-3.10-Administration_guide-en-US/files/prepare-restricted-environment.sh), it seems to me that it requires a mirror registry as a prerequisite.

I managed to get this script to work after:
- deploying a registry in the same cluster (but on another namespace)
- commenting out all the commands in this script where we were relying on the OpenShift internal registry

```
$ ./scripts/prepare-restricted-environment.sh --prod_operator_index "quay.io/rhdh/iib:latest-v4.14-x86_64" --prod_operator_package_name "rhdh" --prod_operator_bundle_name "rhdh-operator" --prod_operator_version "v1.1.0" --my_registry "airgap-registry-airgap-helper-ns.apps.ci-ln-s0zlv6t-72292.origin-ci-int-gce.dev.rhcloud.com"

[...]
[INFO] Creating CatalogSource and ImageContentSourcePolicy
catalogsource.operators.coreos.com/rhdh-disconnected-install created
imagecontentsourcepolicy.operator.openshift.io/rhdh-disconnected-install created
[INFO] Catalog airgap-registry-airgap-helper-ns.apps.ci-ln-s0zlv6t-72292.origin-ci-int-gce.dev.rhcloud.com/rhdh/rhdh-index:v1.1.0 deployed to the airgap-registry-airgap-helper-ns.apps.ci-ln-s0zlv6t-72292.origin-ci-int-gce.dev.rhcloud.com registry.
```

So I think we could maybe change this script so that it requires a registry to be passed. And create a separate helper script that would deploy a registry in the cluster, for folks that do not already have a mirror registry.
@nickboldt WDYT ?

nickboldt: > deploy a registry in the cluster

yeah, that would be great as a default behaviour (and then optionally allow users to bring their own registry)

scripts/prepare-restricted-environment.sh
```diff
@@ -118,6 +118,10 @@ echo "[INFO] Mirroring related images to the $my_registry registry."
 while IFS= read -r line
 do
   public_image=$(echo "${line}" | cut -d '=' -f1)
+  if [[ "$public_image" == registry.redhat.io/rhdh/* ]]; then
```
nickboldt, line 121:
```suggestion
  if [[ $prod_operator_index != "registry.redhat.io/redhat/redhat-operator-index"* ]] || [[ "$public_image" == quay.io/rhdh/* ]]; then
```

Should that be `!=` instead of `==` ?

As this script should work for customers installing GA version (1.1+) to their airgapped environment, pushing them to the quay.io images will allow them to install pre-GA (unsupported) CI builds. We don't want that as the default behaviour.

Maybe we want this to be **optional** behaviour if the `--prod_operator_index "registry.redhat.io/redhat/redhat-operator-index:v4.14"` is from quay instead of from the official rh.reg.io index?

Also what if the `$public_image` is the postgres or rbac image? those are not under `registry.redhat.io/rhdh`.

scripts/prepare-restricted-environment.sh
```diff
@@ -118,6 +118,10 @@ echo "[INFO] Mirroring related images to the $my_registry registry."
...
+    # CI Builds not public yet
+    public_image=${public_image/registry.redhat.io/quay.io}
```
nickboldt, line 123: This will also change references to the rbac and postgres images, so we want to only make this change for our rhdh images:

```suggestion
    public_image=${public_image/registry.redhat.io\/rhdh/quay.io\/rhdh}
```

scripts/prepare-restricted-environment.sh
rm3l: That makes sense - I'll update it accordingly.

scripts/prepare-restricted-environment.sh
rm3l: But rbac and postgresql images are respectively `registry.redhat.io/openshift4/ose-kube-rbac-proxy` and `registry.redhat.io/rhel9/postgresql-15`, so we should not enter this, because of the condition above which catches only images starting with `registry.redhat.io/rhdh`.
Anyway, to be safe, I'll add what you suggest.

scripts/prepare-restricted-environment.sh
rm3l: > As this script should work for customers installing GA version (1.1+) to their airgapped environment, pushing them to the quay.io images will allow them to install pre-GA (unsupported) CI builds. We don't want that as the default behaviour.

Good point.

> Should that be != instead of == ?
> Maybe we want this to be optional behaviour if the --prod_operator_index "registry.redhat.io/redhat/redhat-operator-index:v4.14" is from quay instead of from the official rh.reg.io index?

I might have missed something, but I am not sure I understand the `|| [[ "$public_image" == "quay.io/rhdh/"* ]]` in your suggestion above. We are trying to replace `registry.redhat.io/rhdh` with `quay.io/rhdh`, so this would have no effect if `"$public_image" == "quay.io/rhdh/"*`, no?

So how about something like this:
```bash
if [[ "$prod_operator_index" != registry.redhat.io/redhat/redhat-operator-index* ]] && [[ "$public_image" == registry.redhat.io/rhdh/* ]]; then
...
```
Which means that we would consider replacing `registry.redhat.io/rhdh/*` images only if `prod_operator_index` is not from the official registry.redhat.io..

Additionally, would it make sense to check, using `skopeo inspect`, if the image manifest does exist in the public registry, to decide whether this is (likely) a CI build or not?

So something like this in the end:
```bash
if [[ "$prod_operator_index" != registry.redhat.io/redhat/redhat-operator-index* ]] && [[ "$public_image" == registry.redhat.io/rhdh/* ]]; then
    if ! skopeo inspect "docker://$public_image" &> /dev/null; then
      # likely CI build, which is not public yet
      public_image=${public_image/registry.redhat.io\/rhdh/quay.io\/rhdh}
    fi
  fi
```

.rhdh/scripts/prepare-restricted-environment.sh
```diff
@@ -0,0 +1,329 @@
...
+# SPDX-License-Identifier: EPL-2.0
+#
+
+# Fail on error
+set -e
+
+# example usage:
+# ./prepare-restricted-environment.sh \
+#   --prod_operator_index "registry.redhat.io/redhat/redhat-operator-index:v4.14" \
+#   --prod_operator_package_name "devspaces" \
```
jianrongzhang89, line 17: Why ise devspaces here?

nickboldt: oops, paste error.

nickboldt, line 17:
```suggestion
#   --prod_operator_package_name "rhdh" \
```

scripts/prepare-restricted-environment.sh
nickboldt: skopeo check is a good idea. and your changes to the if-check are definitely better! thanks, see you in a week and a half!

nickboldt:
Changes added to this PR: approved! Please merge when you're happy with it.

rm3l (approved):

Issue #162, opened 2024-01-29
Avoid hardcoding the postgres and backstage images in the controller


kim-tsao:
If the env vars are not specified in the SetupWithManager method, the hardcoded defaults are used but these are upstream values.  We should allow this to be configurable with the default values from downstream.

gazarenkov:
+1 and there are no needs in it since we have our default images configured in non-hardcoded way
https://github.com/janus-idp/operator/blob/main/docs/admin.md

gazarenkov:
https://issues.redhat.com/browse/RHIDP-1190

PR #187, opened 2024-02-06
Avoid hardcoded images

gazarenkov:
## Description

Avoid hardcoded images as stated in https://github.com/janus-idp/operator/issues/162

https://issues.redhat.com/browse/RHIDP-1190

https://github.com/janus-idp/operator/issues/162

## PR acceptance criteria

- [x] Tests

1. make test
2. deploy operator,  kubectl apply -n backstage -f examples/bs1.yaml, make sure the images used are the same as specified in default-config: quay.io/janus-idp/backstage-showcase:latest and quay.io/fedora/postgresql-15:latest)


config/manager/default-config/db-statefulset.yaml
```diff
@@ -29,7 +29,7 @@ spec:
           envFrom:
             - secretRef:
                 name: <POSTGRESQL_SECRET>  # will be replaced with 'backstage-psql-secrets-<cr-name>'
-          image: <RELATED_IMAGE_postgresql> # will be replaced with the actual image
+          image: quay.io/fedora/postgresql-15:latest
```
jianrongzhang89, line 32: Moving the default images to the default configs does not seem to solve the issue as intended. The defaults are still hard-coded although now they are in the configmap.

Also seems to me this does mot make  the operator safer but in opposite: with current implementation if RELATED_IMAGE_postgresql is not provided, the operator will issue an error, while with this PR the defaults get actually

controllers/local_db_statefulset.go
```diff
@@ -219,9 +220,12 @@ func (r *BackstageReconciler) patchLocalDbStatefulSetObj(statefulSet *appsv1.Sta
 
 func (r *BackstageReconciler) setDefaultStatefulSetImage(statefulSet *appsv1.StatefulSet) {
 	visitContainers(&statefulSet.Spec.Template, func(container *v1.Container) {
-		if len(container.Image) == 0 || container.Image == fmt.Sprintf("<%s>", bs.EnvPostGresImage) {
-			container.Image = r.PsqlImage
+		if val, ok := os.LookupEnv(EnvPostgresImage); ok {
```
jianrongzhang89, line 223: This system call get invoked during each reconciliation and can be avoided if the operator reads the env vars during the startup and use the values later.

jianrongzhang89 (changes requested):

config/manager/default-config/db-statefulset.yaml
gazarenkov: No, it is not hardcoded, it is yaml configuration which can be changed by kubectl apply (or so), become configMap 
 and is part of the bundle (along with CSV yaml).
 
 I am not sure I understand your "safer" point. Yes, Admin should specify the default images in the Operator config (as mentioned in Admin guide), if he/she does not it will fail...

controllers/local_db_statefulset.go
gazarenkov: True, and having it reconsiled we can change configuration dynamically, not requiring restarting controller and interrupting service, which is a Kubernetes way I believe.

controllers/backstage_controller.go
```diff
@@ -295,14 +301,13 @@ func (r *BackstageReconciler) labels(meta *v1.ObjectMeta, backstage bs.Backstage
...
-		r.PsqlImage = "quay.io/fedora/postgresql-15:latest"
-		log.Info("Enviroment variable is not set, default is used", bs.EnvPostGresImage, r.PsqlImage)
-	}
 
-	if len(r.BackstageImage) == 0 {
-		r.BackstageImage = "quay.io/janus-idp/backstage-showcase:next"
-		log.Info("Enviroment variable is not set, default is used", bs.EnvBackstageImage, r.BackstageImage)
+	const imageNotSet = "environment variable is not set, default will be used:"
+	if _, ok := os.LookupEnv(EnvPostgresImage); !ok {
+		log.Info(imageNotSet, "", EnvPostgresImage)
```
rm3l, line 307:
```suggestion
		log.Info(imageNotSet, EnvPostgresImage, "")
```

Otherwise, the log message looks weird to me ;)
```
2024-02-09T10:58:24+01:00       INFO    setup   environment variable is not set, default will be used:  {"": "RELATED_IMAGE_postgresql"}
```

controllers/backstage_controller.go
```diff
@@ -295,14 +301,13 @@ func (r *BackstageReconciler) labels(meta *v1.ObjectMeta, backstage bs.Backstage
...
+	}
+	if _, ok := os.LookupEnv(EnvBackstageImage); !ok {
+		log.Info(imageNotSet, "", EnvBackstageImage)
```
rm3l, line 310:
```suggestion
		log.Info(imageNotSet, EnvBackstageImage, "")
```

Otherwise, the log message looks weird to me ;)
```
2024-02-09T10:58:24+01:00       INFO    setup   environment variable is not set, default will be used:  {"": "RELATED_IMAGE_backstage"}
```

controllers/backstage_controller.go
```diff
@@ -295,14 +301,13 @@ func (r *BackstageReconciler) labels(meta *v1.ObjectMeta, backstage bs.Backstage
...
 func (r *BackstageReconciler) SetupWithManager(mgr ctrl.Manager, log logr.Logger) error {
-	if len(r.PsqlImage) == 0 {
-		r.PsqlImage = "quay.io/fedora/postgresql-15:latest"
-		log.Info("Enviroment variable is not set, default is used", bs.EnvPostGresImage, r.PsqlImage)
-	}
 
-	if len(r.BackstageImage) == 0 {
-		r.BackstageImage = "quay.io/janus-idp/backstage-showcase:next"
-		log.Info("Enviroment variable is not set, default is used", bs.EnvBackstageImage, r.BackstageImage)
+	const imageNotSet = "environment variable is not set, default will be used:"
```
rm3l, line 305:
```suggestion
	const imageNotSet = "environment variable is not set, default will be used from config:"
```

controllers/local_db_statefulset.go
```diff
@@ -219,9 +220,12 @@ func (r *BackstageReconciler) patchLocalDbStatefulSetObj(statefulSet *appsv1.Sta
...
+			container.Image = val
 		}
+		//if len(container.Image) == 0 || container.Image == fmt.Sprintf("<%s>", bs.EnvPostgresImage) {
```
rm3l, line 226: Please remove the commented lines below.

controllers/local_db_statefulset.go
rm3l: What do you mean by "change configuration dynamically"? I think the issue here is reading the `RELATED_IMAGE_*` env vars from the system during each reconciliation, which is separate from the configuration.
If any of the env vars are modified, K8s will automatically restart the controller pod, I think. In this case, I agree with @jianrongzhang89 that it would probably make more sense to read the env vars only once at startup to initialize the controller..

examples/janus-cr-with-app-configs.yaml
```diff
@@ -132,8 +133,8 @@ data:
             endpoints:
               /explore-backend-completed:
                 target: 'http://localhost:7017'
-
----
+                
+                ---
```
rm3l, line 137:
```suggestion
---
```

examples/janus-cr-with-app-configs.yaml
```diff
@@ -132,8 +133,8 @@ data:
             endpoints:
               /explore-backend-completed:
                 target: 'http://localhost:7017'
-
----
+
```
rm3l, line 136:
```suggestion

```

rm3l (changes requested):

rm3l (changes requested):
With the changes in the configuration ConfigMap, can you also regenerate the bundle manifests with `make bundle`?
I think [`bundle/manifests/backstage-default-config_v1_configmap.yaml`](https://github.com/janus-idp/operator/blob/main/bundle/manifests/backstage-default-config_v1_configmap.yaml) will need to be updated.

config/manager/default-config/deployment.yaml
```diff
@@ -36,7 +36,8 @@ spec:
           env:
             - name: NPM_CONFIG_USERCONFIG
               value: /opt/app-root/src/.npmrc.dynamic-plugins
-          image: <RELATED_IMAGE_backstage> # will be replaced with the actual image quay.io/janus-idp/backstage-showcase:next
+          #image: <RELATED_IMAGE_backstage> # will be replaced with the actual image quay.io/janus-idp/backstage-showcase:next
+          image: quay.io/janus-idp/backstage-showcase:latest
```
rm3l, line 40: To be consistent, I'd suggest using the same default value for the `RELATED_IMAGE_backstage` env var set in the CSV. It uses a different tag: `quay.io/janus-idp/backstage-showcase:next`. See https://github.com/janus-idp/operator/blob/main/config/manager/manager.yaml#L77

config/manager/default-config/db-statefulset.yaml
gazarenkov: Not sure I understand the point...
Yes, you can change either <RELATED_IMAGE_backstage> or <RELATED_IMAGE_postgresql> independently but my undrstanding is <RELATED_IMAGE_backstage> is applied for both backstage and init container?
Otherwise we need 3 env variables (backstage, init and postgres). Did I miss something?

controllers/local_db_statefulset.go
gazarenkov: I think I do not understand why for simple reading env variable we need to make additional code and pull it through the whole application... What for?

config/manager/default-config/deployment.yaml
gazarenkov: As I remember, some time ago we were suggested to not to use the "next" tag, otherwise I have not prefs.

controllers/local_db_statefulset.go
rm3l: You mentioned that "having it reconsiled we can change configuration dynamically, not requiring restarting controller and interrupting service", but in the case of env vars, we won't be able to change them dynamically without a controller restart.
So instead of reading the same env var (and getting the same value) during each reconciliation (and for all containers), we could avoid that by setting the values at initialization time (as done previously).

rm3l: In that case, maybe you could read the value before visiting all containers, rather than reading it in every container visitor func.

```go
  if val, ok := os.LookupEnv(EnvPostgresImage); ok {
	  visitContainers(&statefulSet.Spec.Template, func(container *v1.Container) {
		  container.Image = val
	  })
  }
```

controllers/backstage_deployment.go
```diff
@@ -291,9 +195,10 @@ func (r *BackstageReconciler) validateAndUpdatePsqlSecretRef(backstage bs.Backst
 }
 
 func (r *BackstageReconciler) setDefaultDeploymentImage(deployment *appsv1.Deployment) {
+
 	visitContainers(&deployment.Spec.Template, func(container *v1.Container) {
-		if len(container.Image) == 0 || container.Image == fmt.Sprintf("<%s>", bs.EnvBackstageImage) {
-			container.Image = r.BackstageImage
+		if val, ok := os.LookupEnv(EnvBackstageImage); ok {
```
rm3l, line 200: Same point here: https://github.com/janus-idp/operator/pull/187/files#r1484578725

config/manager/default-config/deployment.yaml
rm3l: Agreed. So let's use `latest` in the CSV as well.

config/manager/default-config/db-statefulset.yaml
rm3l: My point was that the current implementation on `main` could allow someone to use different images between init and main containers, by not setting the `<RELATED_IMAGE_backstage>` placeholder in the ConfigMap (or setting it for main container only and using different image for init container); and this is not desirable right now.
So I am fine with the changes here :p

controllers/local_db_statefulset.go
gazarenkov: ok

jianrongzhang89: > I think I do not understand why for simple reading env variable we need to make additional code and pull it through the whole application... What for?

This is for code quality. The repeated function calls should be avoided whenever it is possible. This can be easily achieved by storing the data in a global variable.

jianrongzhang89: > True, and having it reconsiled we can change configuration dynamically, not requiring restarting controller and interrupting service, which is a Kubernetes way I believe.

I do not think it supports dynamic configuration change. os.LookupEnv() won't pick up new env variable values unless the pod gets restarted.

gazarenkov: Not sure I understand your point about "repeated function". Could you please show me the place (this code is outdated already)

gazarenkov:
> With the changes in the configuration ConfigMap, can you also regenerate the bundle manifests with `make bundle`? I think [`bundle/manifests/backstage-default-config_v1_configmap.yaml`](https://github.com/janus-idp/operator/blob/main/bundle/manifests/backstage-default-config_v1_configmap.yaml) will need to be updated.

I can but this is a generated file, everyone generating bundle can have actual version of it with "make bundle". 
Would it make more sense to put it to .gitignore instead?

rm3l:
> > With the changes in the configuration ConfigMap, can you also regenerate the bundle manifests with `make bundle`? I think [`bundle/manifests/backstage-default-config_v1_configmap.yaml`](https://github.com/janus-idp/operator/blob/main/bundle/manifests/backstage-default-config_v1_configmap.yaml?rgh-link-date=2024-02-12T12%3A27%3A48Z) will need to be updated.
> 
> I can but this is a generated file, everyone generating bundle can have actual version of it with "make bundle". Would it make more sense to put it to .gitignore instead?

As discussed in the team meeting, having the bundle folder committed to the Git repo (just like other generated files here) makes reviewing it much easier (w/o having to clone the repo and generate the bundle manually). And it seems quite common in other operator repos out there.
So let's commit it, and, as further suggestions to not forget to do so:
- add a note in [developer guide](https://github.com/janus-idp/operator/blob/main/docs/developer.md)
- idea of having a GitHub Action that runs the necessary generation, and either pushes any resulting changes to the PR branch or returns an error

gazarenkov:
> With the changes in the configuration ConfigMap, can you also regenerate the bundle manifests with `make bundle`? I think [`bundle/manifests/backstage-default-config_v1_configmap.yaml`](https://github.com/janus-idp/operator/blob/main/bundle/manifests/backstage-default-config_v1_configmap.yaml) will need to be updated.

ok

bundle/manifests/backstage-operator.clusterserviceversion.yaml
rm3l, line 1: In [L222](https://github.com/janus-idp/operator/pull/187/files#diff-350e6729515e8cd74b357c91611b67dc4e9723daa6863854a252fc6acea032cfR222) of this CSV, `RELATED_IMAGE_backtage`  env var still points to the `next` tag:

```yaml
- name: RELATED_IMAGE_backstage
  value: quay.io/janus-idp/backstage-showcase:next
```

Can we be consistent and use `latest` instead as default?

gazarenkov: fixed

rm3l (approved):
Thanks!

bundle/manifests/backstage-operator.clusterserviceversion.yaml
```diff
@@ -219,7 +220,7 @@ spec:
                 - name: RELATED_IMAGE_postgresql
                   value: quay.io/fedora/postgresql-15:latest
                 - name: RELATED_IMAGE_backstage
-                  value: quay.io/janus-idp/backstage-showcase:next
+                  value: quay.io/janus-idp/backstage-showcase:latest
```
nickboldt, line 223: why use `latest` = 0.1 in main branch... surely you want `next` = 0.2 here?

config/manager/default-config/deployment.yaml
nickboldt: by who? 

why would you want the `main` = 0.2/1.2 operator to install the previous stable `1.1.x` = 0.1/1.1 hub?

Issue #192, opened 2024-02-12
Upgrade project to the latest version of Operator SDK


gazarenkov:
[Deprecation Notice] This version is deprecated.The `go/v3` cannot scaffold projects using kustomize versions v4x+ and cannot fully support Kubernetes 1.25+.It is recommended to upgrade your project to the latest versions available (go/v4).Please, check the migration guide to learn how to upgrade your project

More information: https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.33.0/

nickboldt:
As part of the migration from janus-idp to redhat-developer in https://issues.redhat.com/browse/RHIDP-1021, this will be tracked in https://issues.redhat.com/browse/RHIDP-1292

PR #195, opened 2024-02-12
Auto-push bundle manifests changes to PR branch if needed

rm3l:
## Description
This is to ensure bundle manifests are kept updated when submitting a PR. 

This is a follow-up PR to https://github.com/janus-idp/operator/pull/187#issuecomment-1938816003

## PR acceptance criteria

- [ ] Tests
- [x] Documentation

- ~https://github.com/janus-idp/operator/actions/runs/7878373268?pr=195 is an example of run with outdated bundle manifests~
- ~After applying the resulting patch (attached as a job artifact) or after regenerating the bundle with `make bundle`, the job passes successfully, e.g.: https://github.com/janus-idp/operator/actions/runs/7878650096/job/21497344358?pr=195~

Because the workflow is triggered on `pull_request_target` events, this PR will need to be merged into the base branch to see it in action.

### DIFF
--- a/.github/workflows/pr-bundle-diff-checks.yaml
+++ b/.github/workflows/pr-bundle-diff-checks.yaml
@@ -1,4 +1,4 @@
-# Copyright 2023 The Janus IDP Authors
+# Copyright 2024 The Janus IDP Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/.github/workflows/pr-bundle-diff-checks.yaml
+++ b/.github/workflows/pr-bundle-diff-checks.yaml
@@ -1,0 +1,97 @@
+# Copyright 2023 The Janus IDP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: PR Bundle Manifests Validator
+
+on:
+  # pull_request_target needed to be able to commit and push bundle diffs to external fork PRs.
+  # But we included a manual authorization safeguard to prevent PWN requests. See the 'authorize' job below.
+  pull_request_target:
+    branches: 
+    - main
+    - rhdh-1.[0-9]+
+    - 1.[0-9]+.x
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}
+  cancel-in-progress: true
+
+jobs:
+  authorize:
+    # The 'external' environment is configured with the maintainers team as required reviewers.
+    # All the subsequent jobs in this workflow 'need' this job, which will require manual approval for PRs coming from external forks.
+    # see list of approvers in OWNERS file
+    environment:
+      ${{ (github.event.pull_request.head.repo.full_name == github.repository ||
+      contains(fromJSON('["gazarenkov","jianrongzhang89","kadel","nickboldt","rm3l"]'), github.actor)) && 'internal' || 'external' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: approved
+        run: echo "✓"
+
+  pr-bundle-diff-checks:
+    name: PR Bundle Diff
+    runs-on: ubuntu-latest
+    needs: authorize
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+          ref: ${{ github.event.pull_request.head.ref }}
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Check for outdated bundle
+        id: bundle-diff-checker
+        run: |
+          make bundle
+          git status --porcelain
+          # Since operator-sdk 1.26.0, `make bundle` changes the `createdAt` field from the bundle every time we run it.
+          # The `git diff` below checks if only the createdAt field has changed. If is the only change, it is ignored.
+          # Inspired from https://github.com/operator-framework/operator-sdk/issues/6285#issuecomment-1415350333
+          echo "MANIFESTS_CHANGED=$(if git diff --quiet -I'^    createdAt: ' bundle; then echo "false"; else echo "true"; fi)" >> $GITHUB_OUTPUT
+
+      - name: Commit any manifest changes
+        if: ${{ steps.bundle-diff-checker.outputs.MANIFESTS_CHANGED == 'true' }}
+        run: |
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+          git fetch --prune
+          git pull --rebase --autostash
+          git add -A .
+          git commit \
+            -m "Regenerate bundle manifests" \
+            -m "Co-authored-by: $GITHUB_ACTOR <$GITHUB_ACTOR@users.noreply.github.com>"
+          git push
+
+      - name: Comment on PR if bundle manifests were updated
+        uses: actions/github-script@v7
+        if: ${{ !cancelled() && steps.bundle-diff-checker.outputs.MANIFESTS_CHANGED == 'true' }}
+        continue-on-error: true
+        with:
+          script: |
+            await github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: '⚠️ <b>Files changed in bundle generation<b>!<br/><br/>Those changes to the operator bundle manifests should have been pushed automatically to your PR branch.<br/>You might also need to manually update the [`.rhdh/bundle/manifests/rhdh-operator.csv.yaml`](.rhdh/bundle/manifests/rhdh-operator.csv.yaml) CSV file accordingly.'
+            })
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -81,10 +81,10 @@ jobs:
           retention-days: 5
 
       # gosec needs a "build" stage so connect it to the lint step which we always do
-      - name: build
+      - name: Lint
         run: make lint
 
-      - name: test
+      - name: Test
         # run this stage only if there are changes that match the includes and not the excludes
         if: ${{ env.CHANGES != '' }}
         run: make test
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -109,7 +109,7 @@ jobs:
             })
 
       # gosec needs a "build" stage so connect it to the lint step which we always do
-      - name: Lint
+      - name: build
         run: make lint
 
       - name: Test
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -56,29 +56,57 @@ jobs:
 
       - name: Check for outdated bundle
         id: bundle-diff-checker
+        # Lot of debate (https://github.com/janus-idp/operator/pull/195) whether this should be a warning or an error.
+        # For now, this is will be warning + a comment on the PR if manifests are outdated. This way, PR authors/maintainers can be aware of that fact.
+        continue-on-error: true
         run: |
           make bundle
           git status --porcelain
           # Since operator-sdk 1.26.0, `make bundle` changes the `createdAt` field from the bundle every time we run it.
           # The `git diff` below checks if only the createdAt field has changed. If is the only change, it is ignored.
           # Inspired from https://github.com/operator-framework/operator-sdk/issues/6285#issuecomment-1415350333
-          git diff --quiet -I'^    createdAt: ' bundle || ( echo "===================" && \
-            echo "Files changed in bundle generation." && \
-            echo "Please make sure to regenerate the bundle with 'make bundle' and push the changes." && \
-            echo "Make sure you unset any related env vars like VERSION or IMAGE_TAG_BASE or IMG before running this command, as they may affect the resulting manifests." && \
-            echo "For your convenience, the diff will be attached as a job artifact, so you can easily download and Git-apply it right away." && \
-            echo "You might also need to manually update the CSV in '.rhdh/bundle/manifests/rhdh-operator.csv.yaml' file accordingly." && \
-            echo "===================" && \
+          git diff --quiet -I'^    createdAt: ' bundle || ( echo "::group::WARNINGS" && \
             git --no-pager diff | tee bundle.pr-${{ github.event.number }}.patch && \
+            echo "::warning:: Files changed in bundle generation. Please regenerate the bundle with 'make bundle' and push the changes. For your convenience, the diff is attached as a job artifact, so you can easily download and Git-apply it right away instead. You might also need to manually update the CSV in '.rhdh/bundle/manifests/rhdh-operator.csv.yaml' file accordingly." && \
+            echo "::endgroup::" && \
             exit 1 )
 
       - name: Save bundle diff as patch
+        id: bundle-diff-patch-artifact-upload
         uses: actions/upload-artifact@v4
         if: ${{ !cancelled() && steps.bundle-diff-checker.outcome == 'failure' }}
         with:
           name: bundle-diff-patch
           path: bundle.pr-${{ github.event.number }}.patch
           retention-days: 5
+
+      - name: Comment on PR if bundle manifests are outdated
+        uses: actions/github-script@v7
+        if: ${{ !cancelled() && steps.bundle-diff-checker.outcome == 'failure' }}
+        # TODO(rm3l): this won't work for fork PRs due to permission restrictions. Remove this once this is fixed for fork PRs.
+        continue-on-error: true
+        with:
+          script: |
+            await github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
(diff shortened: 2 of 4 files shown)


.github/workflows/pr.yaml
```diff
@@ -54,11 +54,37 @@ jobs:
         with:
           go-version-file: 'go.mod'
 
+      - name: Check for outdated bundle
```
gazarenkov, line 57: To me it looks like adding complexity for pretty much zero outcome, sorry :) 
This file bundle/manifests files are autogenerated by make bundle, so this can NOT "affect the resulting manifests" and so the message looks confusing.

bundle/manifests/backstage-default-config_v1_configmap.yaml
```diff
@@ -62,9 +62,13 @@ data:
             janus-idp.io/app: backstage-psql-cr1 # placeholder for 'backstage-psql-<cr-name>'
```
gazarenkov, line 62: For my understanding: Why reviewer should spend time reviewing autogenerated files?

gazarenkov (changes requested):

.github/workflows/pr.yaml
rm3l: > To me it looks like adding complexity for pretty much zero outcome, sorry :)

No problem - always open to discussions :)
So the whole value here is to alleviate the load on reviewers, using some automation to streamline the review process as much as we can. This could be part of our PR template checklist, but it is always better if it can be enforced automatically, which is the point here.

> I mean I do not mind of calling extra 'make bundle' if we really need

Sure, we can think of calling `make bundle` because we know that, but if someone else contributes a PR, we would need to think about checking the bundle manifests. Even us, didn't we forget sometimes to regenerate the bundle in most of the PRs where it could have been needed? 
Same point: if you update the go code using the kubebuilder marker comments and don't regenerate the CRDs manifests (I think it already happened recently when updating the RBAC permissions), this would be caught by this validation workflow.

> sending the message about autogenerated by 'make bundle' file looks confusing for me , this can NOT "affect the resulting manifests" for github workflow.

Not sure I understand this statement. You mean the message displayed by this step, like in [this run](https://github.com/janus-idp/operator/actions/runs/7878373268/job/21496491114?pr=195#step:5:49)?
```
Files changed in bundle generation.
Please make sure to regenerate the bundle with 'make bundle' and push the changes.
Make sure you unset any related env vars like VERSION or IMAGE_TAG_BASE or IMG before running this command, as they may affect the resulting manifests.
You might also need to update the CSV in '.rhdh/bundle/manifests/rhdh-operator.csv.yaml' file accordingly.
```
This is a hint to the PR author if they want to call `make bundle` by themselves locally, to pay attention to those variables. By "affect the resulting manifests", I mean that if you had exported an env var called `VERSION` or `IMAGE_TAG_BASE`, the generated CSV will likely contain those values. Please try it by yourself, by calling `export IMAGE_TAG_BASE=my.registry.example.com/janus-idp/operator; make bundle`, and you'll see the result.
But I'm open to a different wording suggestion ;)

The CI is the source of truth, so the diff seen by the CI should be the expected outcome. Initially, I was thinking of making the GitHub job just commit the diff, so that the PR author does not even need to think about it. But for now, erroring out in case of a diff will help educate contributors to think about the resulting manifests (because the CSV in `.rhdh` might also need to be changed). That's why I'm attaching the diff to review and apply it as needed.

bundle/manifests/backstage-default-config_v1_configmap.yaml
rm3l: To me, even if it is generated, this is also part of the code; just like the manifests generated via the [kubebuilder marker comments](https://book.kubebuilder.io/reference/generating-crd). It also seems quite common in the operators world, no? ([opentelemetry-operator](https://github.com/open-telemetry/opentelemetry-operator/tree/main/bundle), [observability-operator](https://github.com/rhobs/observability-operator/tree/main/bundle), [jaeger-operator](https://github.com/jaegertracing/jaeger-operator/tree/main/bundle), [grafana-operator](https://github.com/grafana/grafana-operator/tree/master/bundle), ...)
It also makes it easier to check them with code-scanning tools like SonarCloud, Snyk.

.github/workflows/pr.yaml
jianrongzhang89: Enforcing the validation sounds a good idea to me and it is commonly used by others too.

gazarenkov: What really concerning me is that Developer (in most of the cases triggered by this proposal) may not even have an idea that he/she changed something related.
Like I mentioned: for development I use variable like IMG to test it with my repo/image not touching prod one. It is a valid case, right?
So, let's imagine I have no idea that I did something "affecting the resulting manifests" even not changing source code and this confusing me a lot. It is happened due to particular internal implementation details and I do not think everyone should be aware of it.
TBH I thought you will make GH action provide some automation to change the things silently, I even did not consider any messaging back.

gazarenkov (changes requested):

docs/developer.md
```diff
@@ -95,10 +95,10 @@ make deploy-openshift [IMAGE_TAG_BASE=<your-registry>/backstage-operator]
 ```
 
 ### Modifying the API definitions
-If you are editing the API definitions, generate the manifests such as CRs or CRDs using:
+If you are editing the API definitions, regenerate the bundle using:
 ```sh
-make manifests
+make bundle
```
jianrongzhang89, line 100:
```suggestion
make manifests bundle
```

jianrongzhang89 (changes requested):

.github/workflows/pr.yaml
jianrongzhang89: @gazarenkov Developers should be aware that local builds can generate local-specific manifests, and before they submit a PR, they should re-generate the manifests/bundle without local specific information. This is what I have been doing and it has been working well for me:
unset IMG && unset IMAGE_TAG_BASE && unset VERSION && make manifests bundle

rm3l: > What really concerning me is that Developer (in most of the cases triggered by this proposal) may not even have an idea that he/she changed something related. Like I mentioned: for development I use variable like IMG to test it with my repo/image not touching prod one. It is a valid case, right? So, let's imagine I have no idea that I did something "affecting the resulting manifests" even not changing source code and this confusing me a lot. It is happened due to particular internal implementation details and I do not think everyone should be aware of it. 

If you did not change anything affecting the resulting manifests, then the job would just pass with no action required from you. Otherwise, you would learn that there is something needed. ;)

> TBH I thought you will make GH action provide some automation to change the things silently, I even did not consider any messaging back.

I was about to do it that way, but thought about the downstream CSV under `.rhdh`, which we may need to update as well. So IMO, we'd better enforce those with clear messaging, so that people know.
It is somehow like the lint checks to me. Sometimes, linters might be able to auto-fix some things, but in CI we tend to prefer erroring out, so that developers know for next time.

For reference, I noticed quite a few operators out there have similar validation checks (generating, then checking the git diff). See:
- https://github.com/rhobs/observability-operator/blob/main/.github/workflows/pr-checks.yaml#L60-L62
- https://github.com/os-observability/tempo-operator/blob/main/Makefile#L397-L407
- https://github.com/jaegertracing/jaeger-operator/blob/main/Makefile#L121-L127
- https://github.com/grafana/grafana-operator/blob/master/.github/workflows/pr-validation.yaml#L166-L176

gazarenkov: >Developers should be aware that local builds can generate local-specific manifests,

Even knowing about bundle generation specific, Developers must not keep everything in the head, keeping all the concerns unseparated is not the best way to design.
Failing workflow for fully workable code just because developer did not perform 'make bundle' locally is too confusing especially taking into account that OLM is not the only way of deploying upstream operator (even to Opeshift).

>but thought about the downstream CSV under .rhdh, which we may need to update as well.

Not sure I understand, I thought .rhdh uses manually created manifest, i.e. 'make bundle' makes no diff for it. Did I miss something?

rm3l: > Even knowing about bundle generation specific, Developers must not keep everything in the head, keeping all the concerns unseparated is not the best way to design.
Failing workflow for fully workable code just because developer did not perform 'make bundle' locally is too confusing especially taking into account that OLM is not the only way of deploying upstream operator (even to Openshift).

So that's why enforcing this in CI helps the developer IMO.
Developers are free to iterate no matter how they want locally, but we need to enforce checks after a `git push`, and this includes making sure that it meets the requirements to use the operator in production, e.g.: bundle manifests are consistent with the code changes, potential scanning tools against those manifests, ...
Sure, OLM is not the only way, but it is part of what we ship at the end of the day. So we need to enforce that it can be used consistently there as well.

> Not sure I understand, I thought .rhdh uses manually created manifest, i.e. 'make bundle' makes no diff for it. Did I miss something?
Yes, it is manually updated, but one way to remind us to consider doing so is by highlighting potential changes to the upstream CSV.

gazarenkov: I am OK for warning, letting developer know that it seems the locally generated code does not look the same as it is generated by CI (with possible reasons as mentioned). 

But I disagree of failing workflow for that. Unlike lint (you brought as an analogy):
- developer may not directly deal and care about this aspect (see my example above)
- this has no impact on the end result since CI's "make bundle" will produce exactly the same for synced and un-synced locally generated files

bundle/manifests/backstage-default-config_v1_configmap.yaml
jianrongzhang89: Yes, generated code should also be reviewed. The fact that the code is generated does not mean it works as intended.

jianrongzhang89 (approved):
Sounds good to me - a good practice for consistency and code quality.

.github/workflows/pr.yaml
rm3l: > I am OK for warning, letting developer know that it seems the locally generated code does not look the same as it is generated by CI (with possible reasons as mentioned).
> 
> But I disagree of failing workflow for that. Unlike lint (you brought as an analogy):
> 
> * developer may not directly deal and care about this aspect (see my example above)
> * this has no impact on the end result since CI's "make bundle" will produce exactly the same for synced and un-synced locally generated files

I am not sure a warning in the validate job would be noticed, unfortunately. I think it is preferable to error out by default (since wrong bundle manifests will affect what we ship in the end).
And with the Prow bot, it is possible to override a required check manually if needed (using the `/override` comment, and providing a justification).
Again, if a developer makes changes that don't affect the manifests, nothing will change for them. The validate job will just pass. But if a change is made that changes the manifests, it would be better to know and validate it IMO.

gazarenkov: I understand your point but it does not look friendly and make no sense for developers who will want to try it w/o OLM, which we suggest as a quick start in fact.
For ones who care about it, warning would probably give some hint (TBH, I'd see some documentation emphasising this case) but error is too much b/c it does not impact CI.

bundle/manifests/backstage-default-config_v1_configmap.yaml
gazarenkov: You can review whatever you want but, honestly, how often you review for example zz_generated.deepcopy.go file?

If you tend not trust the tool the best thing is go and check it once (maybe trying theirs tests etc), otherwise it become nightmare.
Thats how separation of concern works I think.

.github/workflows/pr.yaml
rm3l: Ok, let me check if we can display a warning instead.. We'll see how it goes with a warning.

rm3l: @gazarenkov This should be a warning now ([`011654e` (#195)](https://github.com/janus-idp/operator/pull/195/commits/011654e3399aa3d8e7b9300d5f076e1ce8731a2b)), which won't change the outcome of the "PR Validate" job. Let's see how it goes. We can always change it back to an error in the future if it does not help that much.
To make the warning more noticeable to PR authors and reviewers, I was planning to automatically add a comment to the PR, but it does not work currently on fork PRs due to permission restrictions. This can be fixed in a subsequent issue/PR. For now, PR authors/reviewers should think about checking the validation job summary in GH for any warnings.
See https://github.com/janus-idp/operator/actions/runs/7959583931?pr=195 for an example and approve if it looks fine to you.

image

nickboldt: I would agree that **warning are ignored until they're escalated to errors**. Consider the warnings about deprecated versions of nodejs used in GH actions. Until the action fails, no one is going to look at the warnings and update all their action dependencies to newer versions depending on non-EOL nodejs versions. 

image

-- https://github.com/janus-idp/operator/actions/runs/7960349847

nickboldt (changes requested):
if there's a problem with validation, FAIL the build. Warnings are tantamount to :thisisfine: and no one will react.

.github/workflows/pr.yaml
gazarenkov: @nickboldt 
 I may miss something but do not think impact of this PR's warning and warnings about deprecated node.js version are comparable.
 
Let's consider something closer:  https://github.com/janus-idp/operator/issues/192
Would you consider it as "warning are ignored" and "escalate it to error" ?
If so, what concrete action items would you propose?

.github/workflows/pr.yaml
```diff
@@ -56,30 +56,58 @@ jobs:
...
       - name: Save bundle diff as patch
+        id: bundle-diff-patch-artifact-upload
         uses: actions/upload-artifact@v4
         if: ${{ !cancelled() && steps.bundle-diff-checker.outcome == 'failure' }}
         with:
           name: bundle-diff-patch
           path: bundle.pr-${{ github.event.number }}.patch
           retention-days: 5
 
+      - name: Comment on PR if bundle manifests are outdated
```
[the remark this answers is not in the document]
rm3l: > it looks like the major part of pr.yaml is dedicated to the case which even does not impact in the end result (it is only about the file which developer generate or does not generate locally)?

Not sure I completely understand your question. You mean a developer making any changes that potentially affect the manifests (even though they are not aware of that in the beginning)?
In that case, I think it should be the responsibility of the CI to make users aware of that fact so that they can fix up their PR. Doing so locally only might be easy to forget or ignore (on purpose or not).

To clarify, we want to prevent the issue where a developer forgets to regenerate the bundle manifests (and potentially update the one in `.rhdh`), causing a mismatch between the code and those manifests.
Regardless of how (error vs warning) we want to make developers aware of such a potential issue, I think it makes sense to enforce this in CI, not only locally.

.github/workflows/pr.yaml
nickboldt: my point is that if you want people to notice a problem when running a GH action, it needs to FAIL, not WARN.

Warnings are ignored; failures are addressed.

.github/workflows/pr.yaml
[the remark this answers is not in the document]
nickboldt: >  we want to make developers aware of such a potential issue, I think it makes sense to enforce this in CI

+1. 

The only way to block the PR from being merged is to fail the GH action if a mismatch occurs. A warning will never be seen by the PR reviewer.

.github/workflows/pr.yaml
```diff
@@ -54,11 +54,65 @@ jobs:
         with:
           go-version-file: 'go.mod'
 
+      - name: Check for outdated bundle
+        id: bundle-diff-checker
+        # Lot of debate (https://github.com/janus-idp/operator/pull/195) whether this should be a warning or an error.
+        # For now, this is will be warning + a comment on the PR if manifests are outdated. This way, PR authors/maintainers can be aware of that fact.
```
nickboldt, line 60: I guess a comment is at least visible to the PR review process, but if action is required to change the PR before merging, then why not FAIL the action and FORCE the developer to fix their PR, instead of just suggesting what needs doing via a comment?

.github/workflows/pr.yaml
```diff
@@ -54,11 +54,65 @@ jobs:
...
+            echo "::endgroup::" && \
+            exit 1 )
+
+      - name: Save bundle diff as patch
+        id: bundle-diff-patch-artifact-upload
+        uses: actions/upload-artifact@v4
+        if: ${{ !cancelled() && steps.bundle-diff-checker.outcome == 'failure' }}
+        with:
+          name: bundle-diff-patch
+          path: bundle.pr-${{ github.event.number }}.patch
```
nickboldt, line 80: if you can generate a patch, why not just push a commit with this change?

.github/workflows/pr.yaml
```diff
@@ -54,11 +54,65 @@ jobs:
...
+          script: |
+            await github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: '🎉 <b>Bundle manifests are up-to-date!<br/><br/>Please also review the bundle manifests to make sure manual updates to the CSV in [`.rhdh/bundle/manifests/rhdh-operator.csv.yaml`](.rhdh/bundle/manifests/rhdh-operator.csv.yaml) file are not required.'
+            })
+
       # gosec needs a "build" stage so connect it to the lint step which we always do
-      - name: build
```
nickboldt, line 58: see comment above -- `gosec needs a "build" stage so connect it to the lint step which we always do`

.github/workflows/pr.yaml
```diff
@@ -54,11 +54,65 @@ jobs:
...
+      - name: Lint
```
nickboldt, line 112:
```suggestion
      - name: build
```

.github/workflows/pr.yaml
rm3l: Yeah, initially, I was thinking of making the GitHub job just commit the diff, so that the PR author does not even need to think about it (Gennady's concern, I think).
But then I thought that the CSV in .rhdh might also need to be changed, and I am not sure we can handle that automatically.
Now with hindsight, I think it can indeed make sense to:
- automatically commit any diff with this change, which should sort out @gazarenkov's main concern, I think
- comment on the PR (or update the PR template) that the CSV in .rhdh should also be checked

WDYT? @gazarenkov @jianrongzhang89 @nickboldt

rm3l: Okay, I've gone ahead and implemented as proposed above:
- bundle manifests changes will be automatically pushed to the PR branch. Due to permission restrictions when using the built-in `GITHUB_TOKEN` token, this had to be done in a separate workflow triggered on `pull_request_target` events (+ same safeguard permissions we are already using to prevent PWN requests from external PRs)
- comment on the PR + update to the PR template that the CSV in `.rhdh` should also be checked

PTAL

/retitle Auto-generate bundle manifests changes to PR branch if needed

.github/workflows/pr-bundle-diff-checks.yaml
```diff
@@ -0,0 +1,97 @@
+# Copyright 2023 The Janus IDP Authors
```
nickboldt, line 1:
```suggestion
# Copyright 2024 The Janus IDP Authors
```

nickboldt (approved):
seems reasonable, let's run it and see what explodes :D

docs/developer.md
```diff
@@ -95,10 +95,10 @@ make deploy-openshift [IMAGE_TAG_BASE=<your-registry>/backstage-operator]
 ```
 
 ### Modifying the API definitions
-If you are editing the API definitions, generate the manifests such as CRs or CRDs using:
+If you are editing the API definitions, regenerate the manifests and bundle using:
```
gazarenkov, line 98: To cover all the deployment options we offer, I would say something like:
If you are editing the API definitions, make sure you :
- run 'make install' before deploying operator with 'make deploy'
- regenerate the manifests and bundle if you plan to deploy operator with OLM using: ...

PR #200, opened 2024-02-13
Update default VERSION in makefile to 0.0.1-latest

coreydaley:
The `0.0.1` tag does not exist and the `make deploy` will fail if no other `VERSION` is supplied.

- Fixes #_issue_number_

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+VERSION ?= 0.0.1-latest
 
 # Using docker or podman to build and push images
 CONTAINER_ENGINE ?= docker


Makefile
```diff
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 0.0.1
+VERSION ?= 0.0.1-latest
```
rm3l, line 6: Thanks @coreydaley ! But now the tags for the PR images sound a bit confusing to me :/

> PR images are available:
> 
> 1. [quay.io/janus-idp/operator:0.0.1-latest-pr-200-ae0f223](https://quay.io/janus-idp/operator:0.0.1-latest-pr-200-ae0f223)
> 2. [quay.io/janus-idp/operator-bundle:0.0.1-latest-pr-200-ae0f223](https://quay.io/janus-idp/operator-bundle:0.0.1-latest-pr-200-ae0f223)
> 3. [quay.io/janus-idp/operator-catalog:0.0.1-latest-pr-200-ae0f223](https://quay.io/janus-idp/operator-catalog:0.0.1-latest-pr-200-ae0f223)

Also, I think that with the changes here, once this PR is merged, the [`next-container-build`](https://github.com/janus-idp/operator/blob/main/.github/workflows/next-container-build.yaml#L95-L114) workflow would end up creating images like this: `quay.io/janus-idp/operator:0.0.1-latest-latest-$sha` or `quay.io/janus-idp/operator:0.0.1-latest-next-$sha`, which will make it more confusing to me.

Maybe we should just build and push the tag for the base version (`0,0,1`, which would be considered the `latest`) besides the other tags? So I'd suggest keeping the base version as it is, but changing https://github.com/janus-idp/operator/blob/main/.github/workflows/next-container-build.yaml#L88-L114 to push the tag for the base version. WDYT?

rm3l (changes requested):

Makefile
gazarenkov: Good point
I think we should not have "0.0.1" version yet as it is not released yet but we probably need to have "latest" aliased to the last PR's image.
I.e. something like this (for this example):
operator:pr-200-ae0f223 (having 0.0.1 prefix looks redundant to me)
and
operator:latest - pointed to the same image (latest for the time)
And when we release 0.0.1
operator:0.0.1 (plus we may have some alias like "stable" for the latest released)

WDYT?
CC @nickboldt

rm3l: IIRC, `make release-build` does require a semantic version, so with `Major.Minor.Patch` elements:

```
$ make VERSION=pr-200-`git rev-parse --short HEAD` release-build
[...]
Error: invalid command options: pr-200-ae0f223 is not a valid semantic version: No Major.Minor.Patch elements found
```

I think that's why we added the additional labels and build metadata to the version.

gazarenkov: @rm3l I see, I missed this point.
So we can keep it as it is (0.0.1-pr-200-ae0f223) for PR release + tag "latest" for the latest one?

rm3l: So after discussing this, we think it would make sense to:
- set different `VERSION` in the Makefile for [`main`](https://github.com/janus-idp/operator/blob/main/Makefile#L6) and [`1.1.x`](https://github.com/janus-idp/operator/blob/1.1.x/Makefile#L6) branches.
- update the [`next-container-build` workflow](https://github.com/janus-idp/operator/blob/main/.github/workflows/next-container-build.yaml) to also push the tag corresponding to this `VERSION` field

This way, this PR won't be needed, and `make deploy` would work out of the box.

jianrongzhang89: > So after discussing this, we think it would make sense to:
> 
> * set different `VERSION` in the Makefile for [`main`](https://github.com/janus-idp/operator/blob/main/Makefile#L6) and [`1.1.x`](https://github.com/janus-idp/operator/blob/1.1.x/Makefile#L6) branches.
> * update the [`next-container-build` workflow](https://github.com/janus-idp/operator/blob/main/.github/workflows/next-container-build.yaml) to also push the tag corresponding to this `VERSION` field
> 
> This way, this PR won't be needed, and `make deploy` would work out of the box.
This sounds good to me. Further considerations for discussion:
a) In the master branch, set VERSION ?= 1.1.0-dev  (In this way the images from master and release branches will never have conflict).
b) In the release branch 1.1.x, set VERSION ?= 1.1.0

rm3l: @jianrongzhang89 Following the discussion yesterday about [versioning of the CR](https://github.com/janus-idp/operator/pull/201#issuecomment-1946160135), I'm wondering if we should not still keep `0` as major here as well (to make it clear that nothing is not stable version yet - (only RHDH version is 1.1))?
So:
- `0.0.1` as it is in the release branch `1.1.x`
- but `0.0.1-dev` on `main`
WDYT?

jianrongzhang89: @rm3l it makes sense to use major version 0, but I think we should use a minor version instead of a patch version:
0.1.0 as it is in the release branch (and we should rename the branch to 0.1.x to match them),
but 0.1.0-dev on main.

Releases with bugfixes can be released as 0.1.1 etc, releases with new features can be 0.2.0 etc.

rm3l: That makes sense. But I would keep the `1.1.x` release branch name as it is for now, as I'm not sure about the implications on the downstream build pipelines..

jianrongzhang89:
This PR is replaced by https://github.com/janus-idp/operator/pull/207

Issue #241, opened 2024-02-27
Generate operator deployment script


gazarenkov:
To generate deployment script for making it possible to deploy the Operator by kubectl (or oc) like:
`kubectl apply -f <local-file-or-url-to-the-script>`

The goal is to make it possible to generate this script with command like:
`make deployment-script`

rm3l:
I guess this can be closed, now that #242 is merged..

/close

PR #242, opened 2024-02-27
Generate deployment manifest

gazarenkov:
## Description
As stated in https://github.com/janus-idp/operator/issues/241 it adds deployment-script target to the Makefile

https://github.com/janus-idp/operator/issues/241

## PR acceptance criteria

- [x ] Documentation

`> make deployment-script`

It will generate 
`rhdh-operator-<VERSION>.yaml`


Makefile
```diff
@@ -401,3 +401,10 @@ show-img:
 
 show-container-engine:
 	@echo -n $(CONTAINER_ENGINE)
+
+.PHONY: deployment-script
+deployment-script: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
```
rm3l, line 406: Description is bit confusing to me, compared to the `deploy` target.

```
$ make help
[...]
  deploy           Deploy controller to the K8s cluster specified in ~/.kube/config.
  deployment-script  Deploy controller to the K8s cluster specified in ~/.kube/config.
```

Target name too, as a script IMO means something that I can execute; here we'll be applying the manifest, per the official K8s glossary (https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-manifest).


Suggestion:
```suggestion
deployment-manifest: manifests kustomize ## Generate operator manifest in a file ready to be deployed.
```

Makefile
```diff
@@ -401,3 +401,10 @@ show-img:
...
+	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
+	$(KUSTOMIZE) build config/default > rhdh-operator-${VERSION}.yaml
```
rm3l, line 408: Could you also add the `rhdh-operator-*.yaml` to `.gitignore`?
Or I'm wondering if it wouldn't make sense to also commit and push this automatically (to a file like `deploy/rhdh-operator.yaml`)? Just like we are doing with the bundle manifests..

This way, it would become even more straightforward to deploy the operator: `kubectl apply -f https://raw.githubusercontent.com/janus-idp/operator/main/deploy/rhdh-operator.yaml`
WDYT?

rm3l (changes requested):

Makefile
gazarenkov: I would not like to suggest deploying "random" version from git repo in common case (it is good for ones who deal with sources but not sure for everyone else). I better add it to .gitignore as you suggested.

docs/developer.md
```diff
@@ -54,7 +54,16 @@ make run
 
 - Or deploy the controller to the cluster with the image specified by `IMG`:
 ```sh
-make deploy IMG=<your-registry>/backstage-operator:tag
+make deploy [IMG=<your-registry>/backstage-operator:tag]
+```
+
+- To generate deployment script use:
```
rm3l, line 60:
```suggestion
- To generate deployment manifest, use:
```

docs/developer.md
```diff
@@ -54,7 +54,16 @@ make run
...
+```sh
+make deployment-script [IMG=<your-registry>/backstage-operator:tag]
```
rm3l, line 62:
```suggestion
make deployment-manifest [IMG=<your-registry>/backstage-operator:tag]
```

Makefile
rm3l: > I would not like to suggest deploying "random" version from git repo

Well, the point is that this would not be random, but coming from the operator repo itself, so legitimate. Rather than having to deal with cloning the sources..
But anyway, fine for now to just Git-ignore it.
I think, for the future, this file could be generated and attached as a release artifact when we publish a new GH release.

rm3l (approved):

PR #273, opened 2024-03-21
quick fix for https://issues.redhat.com/browse/RHIDP-1701

gazarenkov:
## Description
quick fix for https://issues.redhat.com/browse/RHIDP-1701

## PR acceptance criteria

- [x] Tests

make test


controllers/backstage_deployment.go
```diff
@@ -195,11 +195,23 @@ func (r *BackstageReconciler) validateAndUpdatePsqlSecretRef(backstage bs.Backst
 
 func (r *BackstageReconciler) setDefaultDeploymentImage(deployment *appsv1.Deployment) {
 	if envBackstageImage != "" {
-		visitContainers(&deployment.Spec.Template, func(container *v1.Container) {
-			container.Image = envBackstageImage
-
-		})
+		for _, c := range deployment.Spec.Template.Spec.Containers {
+			if c.Name == _defaultBackstageMainContainerName {
```
rm3l, line 199: I'd suggest replacing based on a placeholder instead. Otherwise, if someone changes the config to use their own images for the main and/or init containers, they will still be replaced here; so same issue.

rm3l: Please note that we have the same issue with the DB StatefulSet as well.

rm3l (changes requested):
Also, I think a test case will be helpful to catch potential regressions on this. Maybe I'm wrong, but I'm not sure this is captured with the existing tests.

controllers/backstage_deployment.go
gazarenkov: Not sure I understand about placeholder. Could you please elaborate?

For the database: Do you see any option for extension it's deployment in the Helm Chart?
(I just would keep this fix as simple as possible)

rm3l: > Not sure I understand about placeholder. Could you please elaborate?

I mean, use by default a placeholder like `<RELATED_IMAGE_backstage>` in the default deployment.yaml manifest, and perform replacements only for places where this placeholder is used explicitly. This way, it is clear that only these placeholders will be replaced, regardless of the container names. 
I remember I saw something like this in a PR ([`83c7187` (#86)](https://github.com/janus-idp/operator/pull/86/commits/83c71872b44faca44af76c3cc201f0914cefdf59)), but it was changed at some point later 😕 

> For the database: Do you see any option for extension it's deployment in the Helm Chart?
(I just would keep this fix as simple as possible)

Maybe not extend it with an extra-container (or why not actually?), but I was thinking of just trying to use a different DB image for example (which is possible in the Helm Chart). Right now, this is not configurable in the CR, so the only way to do this would be via the `rawRuntimeConfig`, but the image would end up being replaced.

rm3l: > > Not sure I understand about placeholder. Could you please elaborate?
> 
> I mean, use by default a placeholder like `<RELATED_IMAGE_backstage>` in the default deployment.yaml manifest, and perform replacements only for places where this placeholder is used explicitly. This way, it is clear that only these placeholders will be replaced, regardless of the container names. I remember I saw something like this in a PR ([`83c7187` (#86)](https://github.com/janus-idp/operator/pull/86/commits/83c71872b44faca44af76c3cc201f0914cefdf59)), but it was changed at some point later 😕
> 

And here for the logic: https://github.com/janus-idp/operator/pull/86/files#diff-8e7e7c9a64394f1ee039417ef32b84b7122580096d39f293a8d976bd2af27f45R308-R314

gazarenkov: Hmm, if you use <RELATED_IMAGE_backstage> in deployment.yaml, what would be the value if I do not put env variable?
> I remember I saw something like this in a PR ([83c7187 (#86)](https://github.com/janus-idp/operator/pull/86/commits/83c71872b44faca44af76c3cc201f0914cefdf59)), but it was changed at some point later

I guess, the changes you mentioned caused by the fact that in this case we have to use somehow hardcoded value which is not good.

>Maybe not extend it with an extra-container (or why not actually?), but I was thinking of just trying to use a different DB image for example (which is possible in the Helm Chart). Right now, this is not configurable in the CR, so the only way to do this would be via the rawRuntimeConfig, but the image would end up being replaced.

OK, Let's move this discussion to the issue?
My point here is to fix concrete problem and so far I can not see it changed anything else but fix it :) . WDYT?

gazarenkov:
Close this and create bug fixes PR for 1.1.x branch.

PR #278, opened 2024-03-27
pin setup-envtest to fix test breakage

kim-tsao:
## Description

Current PR tests are failing due to this error:
`
go: sigs.k8s.io/controller-runtime/tools/setup-envtest@latest (in sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240326031119-8afbb50e0eb2): go.mod:3: invalid go version '1.22.0': must match format 1.23`

as a result of this change https://github.com/kubernetes-sigs/controller-runtime/issues/2720#issuecomment-2016461180
We need to pin to an older commit in order to work around this.

## Which issue(s) does this PR fix or relate to

- Fixes #_issue_number_

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation
- [ ] If the bundle manifests have been updated, make sure to review the [`rhdh-operator.csv.yaml`](../.rhdh/bundle/manifests/rhdh-operator.csv.yaml) file accordingly

PR #279, opened 2024-03-27
overwrite only backstage and init-dynamic-plugins image. fix for https://issues.redhat.com/browse/RHIDP-1701 

gazarenkov:

https://issues.redhat.com/browse/RHIDP-1701 

## PR acceptance criteria

- [x] Tests

### DIFF
--- a/api/v1alpha1/backstage_types.go
+++ b/api/v1alpha1/backstage_types.go
@@ -180,7 +180,7 @@ type Env struct {
 type RuntimeConfig struct {
 	// Name of ConfigMap containing Backstage runtime objects configuration
 	BackstageConfigName string `json:"backstageConfig,omitempty"`
-	// Name of ConfigMap containing LocalDb (P|ostgreSQL) runtime objects configuration
+	// Name of ConfigMap containing LocalDb (PostgreSQL) runtime objects configuration
 	LocalDbConfigName string `json:"localDbConfig,omitempty"`
 }
 
--- a/api/v1alpha1/backstage_types.go
+++ b/api/v1alpha1/backstage_types.go
@@ -180,7 +180,7 @@ type Env struct {
 type RuntimeConfig struct {
 	// Name of ConfigMap containing Backstage runtime objects configuration
 	BackstageConfigName string `json:"backstageConfig,omitempty"`
-	// Name of ConfigMap containing LocalDb (PostgreSQL) runtime objects configuration
+	// Name of ConfigMap containing LocalDb (P|ostgreSQL) runtime objects configuration
 	LocalDbConfigName string `json:"localDbConfig,omitempty"`
 }
 
--- a/config/crd/bases/rhdh.redhat.com_backstages.yaml
+++ b/config/crd/bases/rhdh.redhat.com_backstages.yaml
@@ -292,7 +292,7 @@ spec:
                       configuration
                     type: string
                   localDbConfig:
-                    description: Name of ConfigMap containing LocalDb (P|ostgreSQL)
+                    description: Name of ConfigMap containing LocalDb (PostgreSQL)
                       runtime objects configuration
                     type: string
                 type: object
--- a/config/crd/bases/rhdh.redhat.com_backstages.yaml
+++ b/config/crd/bases/rhdh.redhat.com_backstages.yaml
@@ -292,7 +292,7 @@ spec:
                       configuration
                     type: string
                   localDbConfig:
-                    description: Name of ConfigMap containing LocalDb (PostgreSQL)
+                    description: Name of ConfigMap containing LocalDb (P|ostgreSQL)
                       runtime objects configuration
                     type: string
                 type: object
--- a/controllers/backstage_controller_test.go
+++ b/controllers/backstage_controller_test.go
@@ -1315,7 +1315,9 @@ plugins: []
 			By("Checking that the image was set on all containers in the Pod Spec")
 			visitContainers(&found.Spec.Template, func(container *corev1.Container) {
 				By(fmt.Sprintf("Checking Image in the Backstage Deployment - container: %q", container.Name), func() {
-					Expect(container.Image).Should(Equal(imageName))
+					if container.Image == _defaultBackstageMainContainerName || container.Image == _defaultBackstageInitContainerName {
+						Expect(container.Image).Should(Equal(imageName))
+					}
 				})
 			})
 
--- a/controllers/backstage_controller_test.go
+++ b/controllers/backstage_controller_test.go
@@ -1315,7 +1315,7 @@ plugins: []
 			By("Checking that the image was set on all containers in the Pod Spec")
 			visitContainers(&found.Spec.Template, func(container *corev1.Container) {
 				By(fmt.Sprintf("Checking Image in the Backstage Deployment - container: %q", container.Name), func() {
-					if container.Image == _defaultBackstageMainContainerName || container.Image == _defaultBackstageInitContainerName {
+					if container.Name == _defaultBackstageMainContainerName || container.Name == _defaultBackstageInitContainerName {
 						Expect(container.Image).Should(Equal(imageName))
 					}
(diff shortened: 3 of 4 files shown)


rm3l (commented):
```
test -s /home/runner/work/operator/operator/bin/controller-gen || GOBIN=/home/runner/work/operator/operator/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.11.3
/home/runner/work/operator/operator/bin/controller-gen rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
/home/runner/work/operator/operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
test -s /home/runner/work/operator/operator/bin/goimports || GOBIN=/home/runner/work/operator/operator/bin go install golang.org/x/tools/cmd/goimports@v0.15.0
find . -not -path '*/\.*' -name '*.go' -exec /home/runner/work/operator/operator/bin/goimports -w {} \;
go vet ./...
test -s /home/runner/work/operator/operator/bin/setup-envtest || GOBIN=/home/runner/work/operator/operator/bin go install sigs.k[8](https://github.com/janus-idp/operator/actions/runs/8451633049/job/23150369560?pr=279#step:6:9)s.io/controller-runtime/tools/setup-envtest@latest
go: downloading sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-2024032603111[9](https://github.com/janus-idp/operator/actions/runs/8451633049/job/23150369560?pr=279#step:6:10)-8afbb50e0eb2
go: sigs.k8s.io/controller-runtime/tools/setup-envtest@latest (in sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240326031119-8afbb50e0eb2): go.mod:3: invalid go version '1.22.0': must match format 1.23
make: *** [Makefile:250: /home/runner/work/operator/operator/bin/setup-envtest] Error 1
Error: Process completed with exit code 2.
```

Tests are failing because of `setup-envtest`. This was fixed in #278 (cherry-picked in #280). Could you please rebase your branch onto `1.1.x` ?

api/v1alpha1/backstage_types.go
```diff
@@ -180,7 +180,7 @@ type Env struct {
 type RuntimeConfig struct {
 	// Name of ConfigMap containing Backstage runtime objects configuration
 	BackstageConfigName string `json:"backstageConfig,omitempty"`
-	// Name of ConfigMap containing LocalDb (P|ostgreSQL) runtime objects configuration
+	// Name of ConfigMap containing LocalDb (PostgreSQL) runtime objects configuration
```
rm3l, line 183: This would mean regenerating the bundle manifests.. Since this is not strictly related to RHIDP-1701, can this be done in a separate PR instead?

controllers/backstage_controller_test.go
```diff
@@ -1315,7 +1315,9 @@ plugins: []
 			By("Checking that the image was set on all containers in the Pod Spec")
 			visitContainers(&found.Spec.Template, func(container *corev1.Container) {
 				By(fmt.Sprintf("Checking Image in the Backstage Deployment - container: %q", container.Name), func() {
-					Expect(container.Image).Should(Equal(imageName))
+					if container.Image == _defaultBackstageMainContainerName || container.Image == _defaultBackstageInitContainerName {
```
rm3l, line 1318: container name, you mean, right?

```suggestion
					if container.Name == _defaultBackstageMainContainerName || container.Name == _defaultBackstageInitContainerName {
```

controllers/backstage_deployment.go
```diff
@@ -236,3 +231,11 @@ func getDefaultObjName(backstage bs.Backstage) string {
 func getDefaultDbObjName(backstage bs.Backstage) string {
 	return fmt.Sprintf("backstage-psql-%s", backstage.Name)
 }
+
+func setBackstageImage(deployment *appsv1.Deployment, imageName string) {
+	visitContainers(&deployment.Spec.Template, func(container *v1.Container) {
+		if container.Image == _defaultBackstageMainContainerName || container.Image == _defaultBackstageInitContainerName {
```
rm3l, line 237: container name, you mean, right?

```suggestion
		if container.Name == _defaultBackstageMainContainerName || container.Name == _defaultBackstageInitContainerName {
```

rm3l (changes requested):

controllers/backstage_controller_test.go
rm3l, line 1: Also, what do you think of adding a test for the specific issue we are trying to solve here? i.e., adding a sidecar container via `rawRuntimeConfig` and making sure that the image is not overridden.

gazarenkov: The issue we are solving here is to not to modify the image and, apparently, it should work this way (indeed, it is just a couple of if-s, not a complex logic inside).
So, I would prefer not to extend existing tests b/c in the next version I introduce new integration test framework (this one is in place temporarily until all the cases will be covered with new one, WIP). And I am afraid new, (quite a big chunk of) code will add more mess than bring usefulness, for the time being.
But I will definitely add such tests when we introduce a solution for sidecar containers.
Hope it makes sense.

rm3l: > The issue we are solving here is to not to modify the image

Not sure I understand: the original issue reported in RHIDP-1701 was more around how to include an oauth2-proxy sidecar container with a different image. As a result, it was not possible to do so because the image would be overridden. So it seems the issue we are solving here is indeed to make sure that the image for other containers is not modified. After that, I think the next task would be to document how to add the oauth2-proxy sidecar container to Operator-backed deployments.

> So, I would prefer not to extend existing tests b/c in the next version I introduce new integration test framework (this one is in place temporarily until all the cases will be covered with new one, WIP)

1.1.x branch needs to be maintained too even if `main` is completely different. Adding tests in this branch will also help prevent future regressions, and if cherry-picked to `main` can also help ensure existing tests are not broken.

Actually, I think there is already a test case for `rawRuntimeConfig`: https://github.com/janus-idp/operator/blob/1.1.x/controllers/backstage_controller_test.go#L469C9-L469C65
I was thinking of simply adding an additional Expect there to make sure that the image is not overridden. It should not be that long, and also will help us prevent future regressions (even if the fix is straightforward) ;-)

From my previous review comment in https://github.com/janus-idp/operator/pull/279#discussion_r1541325641, the tests were actually passing but the issue was still there.
If I revert your changes in `controllers/backstage_deployment.go`, the tests would still pass, which is why I was a little bit concerned.

gazarenkov: Ok, I added this test, take a look pls 
> 1.1.x branch needs to be maintained
Ohh, yes, actually maintaining 1.1 is the reason why I proposed this PR :)

config/crd/bases/rhdh.redhat.com_backstages.yaml
```diff
@@ -292,7 +292,7 @@ spec:
                       configuration
                     type: string
                   localDbConfig:
-                    description: Name of ConfigMap containing LocalDb (PostgreSQL)
+                    description: Name of ConfigMap containing LocalDb (P|ostgreSQL)
```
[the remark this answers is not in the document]
gazarenkov: It is "P|ostgreSQL", i.e. the same as it used to be, no?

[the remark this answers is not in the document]
rm3l: I agree it would be consistent with what we have in [`backstage_types.go`](https://github.com/janus-idp/operator/blob/1.1.x/api/v1alpha1/backstage_types.go#L183). But my point was to address this in a separate PR, as it is not related to https://issues.redhat.com/browse/RHIDP-1701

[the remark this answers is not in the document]
gazarenkov: outdated

rm3l (approved):

PR #295, opened 2024-04-04
overwrite only backstage and init-dynamic-plugins image. 

gazarenkov:
## Description

fix for https://issues.redhat.com/browse/RHIDP-1701 

https://issues.redhat.com/browse/RHIDP-1701 

## PR acceptance criteria

- [x] Tests included

### DIFF
--- a/pkg/model/deployment.go
+++ b/pkg/model/deployment.go
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"os"
 
+	"k8s.io/utils/ptr"
+
 	corev1 "k8s.io/api/core/v1"
 
 	bsv1alpha1 "redhat-developer/red-hat-developer-hub-operator/api/v1alpha1"
@@ -77,12 +79,7 @@ func (b *BackstageDeployment) addToModel(model *BackstageModel, _ bsv1alpha1.Bac
 	// override image with env var
 	// [GA] TODO Do we need this feature?
 	if os.Getenv(BackstageImageEnvVar) != "" {
-		b.deployment.Spec.Template.Spec.Containers[0].Image = os.Getenv(BackstageImageEnvVar)
-		// exactly the same image for initContainer and want it to be overriden
-		// the same way as Backstage's one
-		for i := range b.deployment.Spec.Template.Spec.InitContainers {
-			b.deployment.Spec.Template.Spec.InitContainers[i].Image = os.Getenv(BackstageImageEnvVar)
-		}
+		b.setImage(ptr.To(os.Getenv(BackstageImageEnvVar)))
 	}
 
 	return true, nil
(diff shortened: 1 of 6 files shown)


pkg/model/deployment.go
```diff
@@ -155,13 +152,14 @@ func (b *BackstageDeployment) setReplicas(replicas *int32) {
...
+		// this is a workaround for RHDH/Janus configuration
 		// it is not a fact that all the containers should be updated
-		// in general case need something smarter (probably annotation based)
+		// in general case need something smarter
 		// to mark/recognize containers for update
-		VisitContainers(b.podSpec(), func(container *corev1.Container) {
-			container.Image = *image
-		})
+		if len(b.podSpec().InitContainers) > 0 {
+			b.podSpec().InitContainers[0].Image = *image
```
rm3l, line 161: Why does this assume that the first init container is the one to modify? Same in `b.container()` by the way.. I guess this might cause some issues if people prepend additional init containers or containers to the list.
In the fix for 1.1.x (https://github.com/janus-idp/operator/pull/279/files#diff-8e7e7c9a64394f1ee039417ef32b84b7122580096d39f293a8d976bd2af27f45R237), we were leveraging the container names, which I think is a safer alternative, no?

pkg/model/deployment_test.go
```diff
@@ -78,5 +90,6 @@ func TestOverrideBackstageImage(t *testing.T) {
 	assert.NoError(t, err)
```
rm3l, line 90: In L87 above, I'd suggest using [`t.Setenv`](https://pkg.go.dev/testing#T.Setenv) instead of `os.Setenv`. This will guarantee that the env var is restored to its original value after the test.

pkg/model/deployment_test.go
```diff
@@ -78,5 +90,6 @@ func TestOverrideBackstageImage(t *testing.T) {
...
 
 	assert.Equal(t, "dummy", model.backstageDeployment.container().Image)
+	assert.Equal(t, "dummy", model.backstageDeployment.podSpec().InitContainers[0].Image)
```
rm3l, line 93: This test doesn't cover the case where the images for additional user-provided containers should not be modified. This, to me, was the whole point of RHIDP-1701 (issue with the `oauth2-proxy` container), and I think it was tested in https://github.com/janus-idp/operator/pull/279

rm3l (changes requested):

pkg/model/deployment.go
gazarenkov: both equally ugly :) but ok for now, changed

pkg/model/deployment_test.go
gazarenkov: fixed

pkg/model/deployment_test.go
gazarenkov: ok

rm3l (approved):

PR #300, opened 2024-04-09
Run E2E tests against the latest/next available operator images

rm3l:
## Description
I realized that some nightly jobs ([example](https://github.com/janus-idp/operator/actions/runs/8607997036/job/23589525220)) could be skipped, but they should have run actually.
To determine if container images needed to be built upon a push, we were looking at the diff between `HEAD` and `HEAD~1`. The problem is that, if there were some additional changes pushed in, say`HEAD~2`, the job would still be skipped even if those changes were relevant for the container image workflow. And this is also why the E2E tests would be skipped.

In this PR, I'm suggesting we always build container images upon any push to the `main` or release branches, and the E2E would always run against the container images built for the most recent commit.
This should help have meaningful E2E tests that always run against the most recent commit.
And it also makes it easy to re-run and reproduce potential failures with the E2E tests.

Note that this does not change container builds for PRs, which may still be skipped depending on the changes.

&mdash;

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

We'll see this in action after this PR is merged in.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -28,42 +28,18 @@ jobs:
         name: Checkout ${{ matrix.branch }} branch
         run: git switch ${{ matrix.branch }}
 
-      # check changes in this commit for regex include and exclude matches; pipe to an env var
-      - name: Check for changes to build
-        run: |
-          # don't fail if nothing returned by grep
-          set +e 
-          CHANGES="$(git diff --name-only HEAD~1 | \
-            grep -E "workflows/.+-container-build.yaml|Makefile|bundle/|config/|go.mod|go.sum|.+\.go|docker/|\.dockerignore" | \
-            grep -v -E ".+_test.go|/.rhdh/")";
-          echo "Changed files for this commit:"
-          echo "=============================="
-          echo "$CHANGES"
-          echo "=============================="
-          {
-            echo 'CHANGES<<EOF'
-            echo $CHANGES
-            echo EOF
-          } >> "$GITHUB_ENV"
-
       - name: Setup Go
-        # run this stage only if there are changes that match the includes and not the excludes
-        if: ${{ env.CHANGES != '' }}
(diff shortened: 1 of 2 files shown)


nickboldt:
with this change, it looks like you'd be triggering builds now for changes to `.+_test.go|/.rhdh/` which means for a change to the tests or the downstream .rhdh overrides, you'll get a new nightly build of the operator. 

but neither of those regex patterns actually affect the contents of the operator. Are you sure you want extra respins for no-op changes? Seems like a waste of CPU and disk...

Maybe you want to ONLY filter out exclude patterns and build for any change OTHER than those?

rm3l:
/retitle Run E2E tests against the latest/next available operator images

@nickboldt PTAL at the latest changes

.github/workflows/nightly.yaml
```diff
@@ -28,61 +28,26 @@ jobs:
...
-            sleep 2
-            echo ...
-          done
-          echo "... operator image found: ${{ env.OPERATOR_IMAGE }}."
+          latestNext="next"
+          # for main branch, use next tags; for 1.x branches, use :latest tags
+          if [[ $(git rev-parse --abbrev-ref HEAD) != "main" ]]; then
+            latestNext="latest" 
+          fi
+          echo "OPERATOR_IMAGE=quay.io/janus-idp/operator:${latestNext}" >> $GITHUB_ENV
```
nickboldt, line 43: so this workflow doesn't build anything, it just runs tests on the last latest (or next) tagged image? 

why not couple this with the action that creates the new container image (and `:latest` or `:next` tag) ? then you'd get results immediately after each build.

Or is the goal to collect several CI builds into a single daily E2E check, to save resources? If so, +1.

nickboldt (approved):

.github/workflows/nightly.yaml
rm3l: Yup, the goal is indeed to collect several commits and run it on a nightly basis to save resources.

PR #357, opened 2024-05-10
Setup go-imports-organizer and organize Go imports

coreydaley:
This is what it would look like to implement and run the go-imports-organizer on this project. Please note that I have setup the dependencies to be vendored to support not having each user install the go-imports-organizer as a standalone tool

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,6 +105,10 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+.PHONY: fmt
+fmt: ## Run go fmt verifications
+	hack/verify-gofmt.sh
+
 fmt_license: addlicense ## Ensure the license header is set on all files
 	$(ADDLICENSE) -v -f license_header.txt $$(find . -not -path '*/\.*' -name '*.go')
 	$(ADDLICENSE) -v -f license_header.txt $$(find . -name '*ockerfile')
--- a/Makefile
+++ b/Makefile
@@ -105,13 +105,17 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
-.PHONY: fmt
-fmt: goimports ## Format the code using goimports
-	find . -not -path '*/\.*' -name '*.go' -exec $(GOIMPORTS) -w {} \;
-
 fmt_license: addlicense ## Ensure the license header is set on all files
 	$(ADDLICENSE) -v -f license_header.txt $$(find . -not -path '*/\.*' -name '*.go')
 	$(ADDLICENSE) -v -f license_header.txt $$(find . -name '*ockerfile')
+
+.PHONY: imports
+imports: addgoio ## Organize imports in go files using goio.
+	$(GOIO)
+
+.PHONY: verify-imports
+verify-imports: addgoio ## Run import verifications.
+	$(GOIO) -l
 
 .PHONY: gosec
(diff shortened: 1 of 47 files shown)


Makefile
```diff
@@ -415,3 +415,10 @@ show-container-engine:
 	@echo -n $(CONTAINER_ENGINE)
 
 
+imports: ## Organize imports in go files using goio. Example: make imports
```
rm3l, line 418: There is an existing `goimports` target (called as a dependency of `make fmt`). Is it redundant with this one?

Makefile
```diff
@@ -415,3 +415,10 @@ show-container-engine:
...
+	go run ./vendor/github.com/go-imports-organizer/goio
```
rm3l, line 419: I thought we wanted to postpone the decision of vendoring, no? Otherwise, that's a lot of files - makes it even impossible to load the PR files view from my browser.. :-)

Since `go-imports-organizer` can be installed as a standalone binary, can we do a similar thing as what we are doing to download other tools like ginkgo, controller-gen or envtest, i.e., download the binaries to the local `bin` folder if they are missing and use them from there? See https://github.com/janus-idp/operator/blob/main/Makefile#L254-L262

Makefile
coreydaley: `gofmt` is a source code formatter, whereas `goio` is an imports organizer, they perform different functions, so no, it would not be redundant.

Makefile
coreydaley: I can rework the pr to remove the vendoring, though i am pretty sure we will need to add it at some point.

gazarenkov: @coreydaley yes, let's decouple this issue and vendoring please and go with this tool as we do with other (as Armel suggested).

Makefile
rm3l: Yes, but I was talking about `goimports` (from [golang.org/x/tools/cmd/goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports)): https://github.com/janus-idp/operator/blob/main/Makefile#L269-L272
It also organizes the Go import lines (while also formatting the code as `gofmt`) So I was wondering if this should be removed as well..

main.go
```diff
@@ -20,24 +20,21 @@ import (
 	"flag"
 	"os"
 
-	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
-	// to ensure that exec-entrypoint and run can make use of them.
```
rm3l, line 24: Just curious: is this a known limitation of goio to drop such comments from the imports? This might be an issue in some cases, I guess. Fortunately, it didn't drop the `+kubebuilder:scaffold:imports` marker comment.

Makefile
coreydaley: Done.

Makefile
coreydaley: Done.

Makefile
```diff
@@ -106,13 +106,21 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 .PHONY: fmt
-fmt: goimports ## Format the code using goimports
-	find . -not -path '*/\.*' -name '*.go' -exec $(GOIMPORTS) -w {} \;
+fmt: ## Run go fmt verifications
+	hack/verify-gofmt.sh
```
rm3l, line 110: I used to run `make fmt` to have the code formatted, but with this, it would just verify and exit, which is kinda surprising to me. If we are just calling `./hack/verify-gofmt.sh`, I would suggest renaming this Make target into `verify-fmt` to make the intent clearer..

I noticed that `goio` also formats the code as well, besides organizing the imports. So maybe this target could just be this instead:
```
fmt: imports ## Format the code using goio
```

Then a separate `verify-fmt` target that would run `hack/verify-gofmt.sh`.

WDYT?

gazarenkov: Do we need fmt verification at all?

gazarenkov:
Postpone for 1.3

Makefile
rm3l: @coreydaley WDYT?

rm3l (commented):
@coreydaley Could you please rebase your branch and fix the conflicts?

coreydaley:
@rm3l This pull request has just been rebased.

Issue #360, opened 2024-05-14
Setting custom `spec.application.route.host` not working using the RHDH Operator on OpenShift, due to inconsistent permissions between downstream and upstream CSVs


rm3l:
/kind bug

## What versions of software are you using?

- RHDH operator next version (image: `registry.redhat.io/rhdh/rhdh-rhel9-operator@sha256:f9e4e29b935cae26df62191e59f8240cddcc160d0ed29efe9d7d6f9ac549bc8e`)
- or RHDH operator latest released version (image: `registry.redhat.io/rhdh/rhdh-rhel9-operator@sha256:9bea7eabdea44342248dfd6091a8f5c3b6e65884c916b3c2a073a4a64481aa6f`)
- OpenShift 4.15.0

## What did you do exactly?

Against an OpenShift cluster:

```shell
# Similar issue when installing the released operator from the OpenShift OperatorHub
.rhdh/scripts/install-rhdh-catalog-source.sh --next --install-operator rhdh

cat <<EOF | oc apply -f -
apiVersion: rhdh.redhat.com/v1alpha1
kind: Backstage
metadata:
  name: test-bs-route
spec:
  application:
    route:
      host: test-bs-route.example.com
EOF
```

## Actual behavior

The CR status is `DeployFailed`, and not all resources are created.

### Using the upcoming 1.2 (--next)
The error message is `failed to patch object *v1.Deployment: deployments.apps "backstage-test-bs-route" is forbidden: User "system:serviceaccount:rhdh-operator:rhdh-operator" cannot patch resource "deployments" in API group "apps" in the namespace "my-ns"`, which indicates a missing role for the operator service account.

### Using the released 1.1.2 (--latest)
The error message is `failed to deploy Backstage Route: Route.route.openshift.io "backstage-test-bs-route" is invalid: spec.host: Forbidden: you do not have permission to set the host field of the route`, which also indicates a missing role for the operator service account.

## Expected behavior

The downstream RHDH operator should reconcile successfully without any errors, and the Route created by the Operator should have the specified host set.

It works as expected when running the Backstage operator (using `make deploy` or `make run`), not the downstream RHDH bundle. This indicates an issue with out-of-sync RBAC permissions between both bundles, as already caught in https://github.com/janus-idp/operator/pull/351#discussion_r1599790206

We should make sure to keep those permissions in sync.

## Any logs, error output, etc?

- Using the upcoming 1.2 (--next):

```
$ oc get route
No resources found in my-ns namespace.

[...]

PR #361, opened 2024-05-14
Fix the RBAC permissions of the Operator in the RHDH CSV

rm3l:
## Description
There was a difference in the RBAC permissions of the operator between the CSV used for RHDH (https://github.com/janus-idp/operator/blob/main/.rhdh/bundle/manifests/rhdh-operator.csv.yaml) and the upstream one (https://github.com/janus-idp/operator/blob/main/bundle/manifests/backstage-operator.clusterserviceversion.yaml).
This causes issues like the one reported in #360.

- Fixes https://github.com/janus-idp/operator/issues/360
- Fixes https://issues.redhat.com/browse/RHIDP-2325

## PR acceptance criteria

- [x] Tests
- [ ] Documentation

Diff should only be about the service account name:

```diff
$ diff -U 1 \
    <(yq '.spec.install.spec.clusterPermissions' bundle/manifests/backstage-operator.clusterserviceversion.yaml) \
    <(yq '.spec.install.spec.clusterPermissions' .rhdh/bundle/manifests/rhdh-operator.csv.yaml)

--- /proc/self/fd/11    2024-05-14 16:28:32.538242718 +0200
+++ /proc/self/fd/12    2024-05-14 16:28:32.539242725 +0200
@@ -107,2 +107,2 @@
         - create
-  serviceAccountName: backstage-controller-manager
+  serviceAccountName: rhdh-operator
```

We can retry the repro steps depicted in https://github.com/janus-idp/operator/issues/360 after a new downstream build is done (after this PR merged).

### DIFF
--- a/.github/workflows/pr-bundle-diff-checks.yaml
+++ b/.github/workflows/pr-bundle-diff-checks.yaml
@@ -97,3 +97,10 @@ jobs:
               repo: context.repo.repo,
               body: '⚠️ <b>Files changed in bundle generation!</b><br/><br/>Those changes to the operator bundle manifests should have been pushed automatically to your PR branch.<br/>You might also need to manually update the [`.rhdh/bundle/manifests/rhdh-operator.csv.yaml`](${{ env.GH_BLOB_VIEWER_BASE_URL }}/.rhdh/bundle/manifests/rhdh-operator.csv.yaml) CSV file accordingly.'
             })
+
+      - name: Check if the CSV for RHDH needs to be updated
+        run: |
+          echo "Checking that the RBAC roles of the downstream RHDH operator are not out of sync with the upstream CSV..."
+          diff -U 1 \
+            <(yq '.spec.install.spec.clusterPermissions' bundle/manifests/backstage-operator.clusterserviceversion.yaml | grep -v 'serviceAccountName: ') \
+            <(yq '.spec.install.spec.clusterPermissions' .rhdh/bundle/manifests/rhdh-operator.csv.yaml | grep -v 'serviceAccountName: ')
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-bundle-diff-checks.yaml
```diff
@@ -97,3 +97,10 @@ jobs:
               repo: context.repo.repo,
               body: '⚠️ <b>Files changed in bundle generation!</b><br/><br/>Those changes to the operator bundle manifests should have been pushed automatically to your PR branch.<br/>You might also need to manually update the [`.rhdh/bundle/manifests/rhdh-operator.csv.yaml`](${{ env.GH_BLOB_VIEWER_BASE_URL }}/.rhdh/bundle/manifests/rhdh-operator.csv.yaml) CSV file accordingly.'
             })
+
+      - name: Check if the CSV for RHDH needs to be updated
+        run: |
+          echo "Checking that the RBAC roles of the downstream RHDH operator are not out of sync with the upstream CSV..."
+          diff -U 1 \
+            <(yq '.spec.install.spec.clusterPermissions' bundle/manifests/backstage-operator.clusterserviceversion.yaml | grep -v 'serviceAccountName: ') \
+            <(yq '.spec.install.spec.clusterPermissions' .rhdh/bundle/manifests/rhdh-operator.csv.yaml | grep -v 'serviceAccountName: '
```
gazarenkov, line 106: Do you have an idea what prevents us to auto generate the CSV file for downstream instead?

rm3l: Good question. Not sure, but I guess it should be possible to rely on Kustomize to generate this file for RHDH based on a set of patches we would use.
Let's file a separate issue for this. cc @nickboldt for any thoughts or concerns.

gazarenkov (approved):

PR #363, opened 2024-05-21
[RHIDP-2246] Fix reconciliation issue when referenced ConfigMap/Secret has a dot (`.`) in its name

rm3l:
## Description
This PR fixes an issue when a ConfigMap or Secret is referenced in the CR and contains a dot (`.`) in its name. The operator would use the ConfigMap/Secret name as is as the volume name in the Deployment Spec, which could be reported as invalid by Kubernetes.
This PR makes sure to sanitize volume names, such that they are RFC 1123-compliant, as depicted in https://kubernetes.io/docs/concepts/overview/working-with-objects/names/.

- Fixes https://issues.redhat.com/browse/RHIDP-2246

## PR acceptance criteria

- [x] Tests
- [ ] Documentation

See the updated example in [examples/rhdh-cr-with-app-configs.yaml](https://github.com/rm3l/janus-idp-operator/blob/RHIDP-2246--cannot-mount-a-secret-configmap-with-a-.-in-its-name/examples/rhdh-cr-with-app-configs.yaml).
Prior to the changes here, this example would not be reconciled properly. Now it works as expected.

### DIFF
--- a/examples/rhdh-cr-with-app-configs.yaml
+++ b/examples/rhdh-cr-with-app-configs.yaml
@@ -19,14 +19,9 @@ spec:
       mountPath: /tmp/my-extra-files
       configMaps:
         - name: "my-backstage-extra-files-cm1"
-        # RHIDP-2246: mounting secret or configmap with a '.' in the name
-        - name: "my-extra-file-cm.txt"
       secrets:
         - name: "my-backstage-extra-files-secret1"
           key: secret_file1.txt
-        # RHIDP-2246: mounting secret or configmap with a '.' in the name
-        - name: "my-extra-file-secret.txt"
-          key: extra-file-secret.1
     extraEnvs:
       envs:
         - name: GITHUB_ORG
@@ -191,20 +186,6 @@ data:
 
 ---
 apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: my-extra-file-cm.txt
-data:
-  my-extra-file-cm-1.txt: |
-    # From ConfigMap with . in its name
-    Hello World
-  my-extra-file-cm-2.txt: |
-    # From ConfigMap with . in its name
-    some:
-      yaml: definition
-
----
-apiVersion: v1
 kind: Secret
 metadata:
   name: my-backstage-extra-files-secret1
@@ -224,15 +205,3 @@ stringData:
     webhookSecret: someWebhookSecret
     privateKey: |
       SomeRsaPrivateKey
-
----
-apiVersion: v1
-kind: Secret
-metadata:
-  name: my-extra-file-secret.txt
-stringData:
-  extra-file-secret.1: |
-    # From Secret with . in its name
-    <settings>
-      <somefield>some value</somefield>
-    </setting>
--- a/examples/rhdh-cr-with-app-configs.yaml
+++ b/examples/rhdh-cr-with-app-configs.yaml
@@ -19,9 +19,14 @@ spec:
       mountPath: /tmp/my-extra-files
       configMaps:
         - name: "my-backstage-extra-files-cm1"
+        # RHIDP-2246: mounting secret or configmap with a '.' in the name
+        - name: "my-extra-file-cm.txt"
       secrets:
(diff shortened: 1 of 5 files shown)


pkg/utils/utils.go
```diff
@@ -53,7 +57,7 @@ func GenerateRuntimeObjectName(backstageCRName string, objectType string) string
 
 // GenerateVolumeNameFromCmOrSecret generates volume name for mounting ConfigMap or Secret
 func GenerateVolumeNameFromCmOrSecret(cmOrSecretName string) string {
-	return cmOrSecretName
+	return ToKubernetesResourceName(cmOrSecretName)
```
gazarenkov, line 60: Well it is hardly something we can call "Kubernetes name", what we need here is simplier conversion from RFC 1123 DNS subdomain name to RFC 1123 label. Which means removing/replacing dots and truncating the string to 63 because cmOrSecretName is already valid RFC 1123 DNS subdomain name.
Also, there is unaddressed problem in  case if transformed string will fit other (the same) name. It is very rare case IMO but still.
WDYT?

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"regexp"
+	"strings"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+func TestToKubernetesResourceName(t *testing.T) {
```
gazarenkov, line 11: Most (if not all) of the tests are hardly realistic scenarios because K8s just will not allow non - RFC 1123 DNS subdomain name as a CM or Secret name (which we use to construct Volume name) ?

examples/rhdh-cr-with-app-configs.yaml
```diff
@@ -19,9 +19,14 @@ spec:
       mountPath: /tmp/my-extra-files
       configMaps:
         - name: "my-backstage-extra-files-cm1"
+        # RHIDP-2246: mounting secret or configmap with a '.' in the name
```
gazarenkov, line 22: Do we really need to maintain the code for this case in, e2e, integration and unit tests?
Can not it be simply covered with unit tests only instead to simplify maintainability?

gazarenkov (changes requested):

pkg/utils/utils_test.go
rm3l: I've simplified the unit test cases there to focus on what's makes sense to me.

pkg/utils/utils.go
rm3l: > Well it is hardly something we can call "Kubernetes name"
> what we need here is simplier conversion from RFC 1123 DNS subdomain name to RFC 1123 label. Which means removing/replacing dots and truncating the string to 63 because cmOrSecretName is already valid RFC 1123 DNS subdomain name.
>

As discussed together, I renamed this to `ToRFC1123Label`. See [`aedd2d1` (#363)](https://github.com/janus-idp/operator/pull/363/commits/aedd2d14ad22a901899e9d18c40af252ce890445).

There are some edge cases requiring more than a simple replacement, especially with truncation, where we might end up with invalid names, even if the original cmOrSecretName is valid. Example with a valid CM/Secret name like `ppxkgq-df-yyatvyrgjtwivunibicne-bvyyotvonbrtfv-awylmrez.ksvqjw--zxpgdi`.

> Also, there is unaddressed problem in case if transformed string will fit other (the same) name. It is very rare case IMO but still.

As discussed also, let me know if you find a way to handle this rare case. Otherwise, this can be addressed in a separate issue/PR.

examples/rhdh-cr-with-app-configs.yaml
rm3l: As discussed, I think it makes sense to keep at least the unit and integration tests. I've removed the changes to this example.
[`aedd2d1` (#363)](https://github.com/janus-idp/operator/pull/363/commits/aedd2d14ad22a901899e9d18c40af252ce890445)

gazarenkov (approved):

PR #366, opened 2024-05-27
chore: set `NODE_OPTIONS=--no-node-snapshot` env variable by default

Zaperex:
## Description

Now that the RHDH image runs on NodeJS 20, the `NODE_OPTIONS=--no-node-snapshot` environmental variable is required to utilize the backstage scaffolder due to requirements from `isolated-vm`.

References: https://github.com/backstage/backstage/issues/20661
https://github.com/laverdet/isolated-vm?tab=readme-ov-file#requirements

- Fixes [RHIDP-2436](https://issues.redhat.com/browse/RHIDP-2436)

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/config/manager/default-config/deployment.yaml
+++ b/config/manager/default-config/deployment.yaml
@@ -99,6 +99,8 @@ spec:
           env:
             - name: APP_CONFIG_backend_listen_port
               value: "7007"
+            - name: NODE_OPTIONS
+              value: '--no-node-snapshot'
           volumeMounts:
             - mountPath: /opt/app-root/src/dynamic-plugins-root
               name: dynamic-plugins-root
--- a/bundle/manifests/backstage-default-config_v1_configmap.yaml
+++ b/bundle/manifests/backstage-default-config_v1_configmap.yaml
@@ -260,6 +260,8 @@ data:
               env:
                 - name: APP_CONFIG_backend_listen_port
                   value: "7007"
+                - name: NODE_OPTIONS
+                  value: '--no-node-snapshot'
               volumeMounts:
                 - mountPath: /opt/app-root/src/dynamic-plugins-root
                   name: dynamic-plugins-root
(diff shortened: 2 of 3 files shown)


config/manager/default-config/deployment.yaml
```diff
@@ -99,6 +99,8 @@ spec:
           env:
             - name: APP_CONFIG_backend_listen_port
               value: "7007"
+            - name: NODE_OPTIONS
+              value: '--no-node-snapshot'
```
rm3l, line 103: Just curious, why this is not defined in the Dockerfile (https://github.com/janus-idp/backstage-showcase/blob/main/docker/Dockerfile and https://github.com/janus-idp/backstage-showcase/blob/main/.rhdh/docker/Dockerfile)? This would avoid having to set it in both the operator and the Helm Chart (https://github.com/redhat-developer/rhdh-chart/pull/14).

rm3l: Also, could this option have any effect on previous versions of RHDH not running on Node 20?

nickboldt: > Also, could this option have any effect on previous versions of RHDH not running on Node 20?

Unlikely, as the chart/operator will deploy a new RHDH container image, which bundles newer plugins.

nickboldt: I'm ok with this change but I agree, if we can fix it in the hub container instead of BOTH the chart and operator, that would seem cleaner.

Zaperex: I'll make a PR in the showcase to update the dockerfile.

Zaperex: https://github.com/janus-idp/backstage-showcase/pull/1277

rm3l:
/hold

This PR might not be needed. Superseded by https://github.com/janus-idp/backstage-showcase/pull/1277

PR #368, opened 2024-05-28
external db documentation

gazarenkov:
## Description

Added external DB configuration documentation

https://issues.redhat.com/browse/RHIDP-2016

### DIFF
--- a/docs/external-db.md
+++ b/docs/external-db.md
@@ -58,7 +58,7 @@ stringData:
 - add **<crt-secret>** to **spec.application.extraFiles.secrets**, so, as for example below **postgres-crt.pem** file will be mounted to Backstage container at **spec.application.extraFiles.mountPath** directory:   
 - add **<cred-secret>** to **spec.application.extraEnvs.secrets**, so all the data's entries will be injected to Backstage container as environment variables.
 
-**NOTE:** environment variables listed in **<cred-secret>** file work with default Operator configuration. If it is changed on or Raw configuration, you have to re-configure it accordingly.
+**NOTE:** environment variables listed in **<cred-secret>** file work with default Operator configuration. If it is changed on default or raw configuration, you have to re-configure it accordingly.
 
 ````
 cat <<EOF | kubectl -n <your-namespace> create -f -
--- a/docs/external-db.md
+++ b/docs/external-db.md
@@ -64,7 +64,7 @@ stringData:
 
 ````yaml
 cat <<EOF | kubectl -n <your-namespace> create -f -
-apiVersion: v1
+apiVersion: rhdh.redhat.com/v1alpha1
 kind: Backstage
 metadata:
  name: <backstage-instance-name>
--- a/docs/external-db.md
+++ b/docs/external-db.md
@@ -40,7 +40,7 @@ EOF
 ### Create secret with certificate(s):
 (omit this step if you do not need TLS connection, maybe for testing purpose)
 
-````
+````yaml
 cat <<EOF | kubectl -n <your-namespace> create -f -
 apiVersion: v1
 kind: Secret
--- a/docs/external-db.md
+++ b/docs/external-db.md
@@ -20,7 +20,7 @@ You can find configuration guidelines for:
 - [Azure Database PostgreSQL](#azure-db-postgresql)
 
 ### Create secret with PostgreSQL connection properties:
-````
+````yaml
 cat <<EOF | kubectl -n <your-namespace> create -f -
 apiVersion: v1
 kind: Secret
--- a/docs/external-db.md
+++ b/docs/external-db.md
@@ -1,0 +1,159 @@
+# External DB integration
+
+Backstage hosts the data in the [PostgreSQL database](https://backstage.io/docs/getting-started/config/database/).
+By default, Operator creates and manages local instance of PostgreSQL in the same as Backstage deployment namespace but it also allows to switch this off and configure external instance instead.
+Usually, external connection requires more security, so, this instruction includes steps to configure SSL/TLS.
+
+### Configure your external PostgreSQL instance
+As a prerequisite you have to know:
+- **db-host** - your PostgreSQL instance DNS or IP address 
+- **db-port** - your PostgreSQL instance port number (usually 5432)
+- **username** - to connect to your PostgreSQL instance
+- **password** - to connect to your PostgreSQL instance
+
+In addition, to get your database connection secured with SSL/TLS you also need certificates in a form of PEM file. 
+
+You can find configuration guidelines for:
+- [AWS RDS PostgreSQL](#aws-rds-postgresql)
+- [Azure Database PostgreSQL](#azure-db-postgresql)
+
+### Create secret with PostgreSQL connection properties:
+````
+cat <<EOF | kubectl -n <your-namespace> create -f -
+apiVersion: v1
+kind: Secret
+metadata:
+ name: <cred-secret-name>
+type: Opaque
+stringData:
+ POSTGRES_PASSWORD: <password>
+ POSTGRES_PORT: <db-port>
+ POSTGRES_USER: <username>
+ POSTGRES_HOST: <db-host>
+ PGSSLMODE: require #  for TLS connection
+ NODE_EXTRA_CA_CERTS: <abs-path-to-pem-file> # for TLS connection, e.g. /opt/app-root/src/postgres-crt.pem
+EOF
+````
+
+### Create secret with certificate(s):
+(omit this step if you do not need TLS connection, maybe for testing purpose)
+
+````
+cat <<EOF | kubectl -n <your-namespace> create -f -
+apiVersion: v1
+kind: Secret
+metadata:
+ name: <crt-secret>
+type: Opaque
+stringData:
+ postgres-crt.pem: |-
+   -----BEGIN CERTIFICATE-----
+   MIIFqDCCA5CgAwIBAgIQHtOXCV/YtLNHcB6qvn9FszANBgkqhkiG9w0BAQwFADBl
+   ... 
+````
+
+### Create Backstage Custom Resource:
+
+- disable creating local PostgreSQL instance with **spec.database.enableLocalDb: false**
+- add **<crt-secret>** to **spec.application.extraFiles.secrets**, so, as for example below **postgres-crt.pem** file will be mounted to Backstage container at **spec.application.extraFiles.mountPath** directory:   
+- add **<cred-secret>** to **spec.application.extraEnvs.secrets**, so all the data's entries will be injected to Backstage container as environment variables.
+
+**NOTE:** environment variables listed in **<cred-secret>** file work with default Operator configuration. If it is changed on or Raw configuration, you have to re-configure it accordingly.
+
+````
+cat <<EOF | kubectl -n <your-namespace> create -f -
+apiVersion: v1
+kind: Secret
+metadata:
+ name: <backstage-instance-name>
+spec:
+ database:
+   enableLocalDb: false 
+ application: 
+    extraFiles:
+     mountPath: <path> # e g /opt/app-root/src
+     secrets:
+       - name: <crt-secret> 
+         key: postgres-crt.pem # key name as in <crt-secret> Secret
+    extraEnvs:
+      secrets:
+        - name: <cred-secret>  
+````
+
+## External PostgreSQL types
+
+### AWS RDS PostgreSQL
+(Tested on PGSQL 15)
+
+#### Prerequisites
+- An AWS account with active subscription and PostgreSQL instance on [Amazon RDS for PostgreSQL](https://aws.amazon.com/rds/postgresql/) 
+- (Optionally) Pgsql client installed to check your database connections 
+
+#### Preparation
+- (Optionally) Check your Database connection:
+
+````
+psql -h <db-host> -p <db-port> -U <username>
+````
+
+- Enter the <password> and output should be something like:
+
+````
+SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, compression: off)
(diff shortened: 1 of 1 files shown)


gustavolira (approved):

docs/external-db.md
```diff
@@ -0,0 +1,159 @@
+# External DB integration
+
+Backstage hosts the data in the [PostgreSQL database](https://backstage.io/docs/getting-started/config/database/).
+By default, Operator creates and manages local instance of PostgreSQL in the same as Backstage deployment namespace but it also allows to switch this off and configure external instance instead.
```
kim-tsao, line 4: This talks about the operator but do we need to document this for the helm chart too?  I imagine most of it will be similar

docs/external-db.md
```diff
@@ -0,0 +1,159 @@
...
+
+````
+cat <<EOF | kubectl -n <your-namespace> create -f -
+apiVersion: v1
+kind: Secret
+metadata:
+ name: <crt-secret>
+type: Opaque
+stringData:
+ postgres-crt.pem: |-
```
rm3l, line 49: Rather than copy-pasting the content of the file, how about using a command like `kubectl create secret generic NAME --from-file=...` ([ref](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_create/kubectl_create_secret_generic/)) to create a secret from a file? I think it is more user-friendly.

docs/external-db.md
```diff
@@ -0,0 +1,159 @@
...
+   -----BEGIN CERTIFICATE-----
+   MIIFqDCCA5CgAwIBAgIQHtOXCV/YtLNHcB6qvn9FszANBgkqhkiG9w0BAQwFADBl
+   ... 
+````
+
+### Create Backstage Custom Resource:
+
+- disable creating local PostgreSQL instance with **spec.database.enableLocalDb: false**
+- add **<crt-secret>** to **spec.application.extraFiles.secrets**, so, as for example below **postgres-crt.pem** file will be mounted to Backstage container at **spec.application.extraFiles.mountPath** directory:   
+- add **<cred-secret>** to **spec.application.extraEnvs.secrets**, so all the data's entries will be injected to Backstage container as environment variables.
```
rm3l, line 59: I know that it would achieve the same result, but how about using the [`spec.database.authSecretName` field](https://github.com/janus-idp/operator/blob/main/api/v1alpha1/backstage_types.go#L68) instead? As the name suggests, it is the secret for database authentication, so I would expect this info to be defined there.

docs/external-db.md
```diff
@@ -0,0 +1,159 @@
...
+        - name: <cred-secret>  
+````
+
+## External PostgreSQL types
+
+### AWS RDS PostgreSQL
+(Tested on PGSQL 15)
+
+#### Prerequisites
+- An AWS account with active subscription and PostgreSQL instance on [Amazon RDS for PostgreSQL](https://aws.amazon.com/rds/postgresql/)
```
rm3l, line 89:
```suggestion
- An AWS account with an active subscription and a PostgreSQL instance on [Amazon RDS for PostgreSQL](https://aws.amazon.com/rds/postgresql/) 
```

docs/external-db.md
```diff
@@ -0,0 +1,159 @@
...
+
+**NOTE:**  AWS RDS **enforces** connecting your client applications using Transport Layer Security (TLS) starting from PGSQL v 15 . You can disable it adding Parameter Group and setting **rds.force-ssl=0**
+
+- Use this PEM file (**postgres-crt.pem**) as a data for the  **<crt-secret>** Secret above.
+
+### Azure DB PostgreSQL
+(Tested on PGSQL 15)
+
+#### Prerequisites
+- An [Azure](https://azure.microsoft.com/) account with active subscription and [Azure Database for PostgreSQL - Flexible Server instance](https://learn.microsoft.com/en-gb/azure/postgresql/flexible-server/overview).
```
rm3l, line 121:
```suggestion
- An [Azure](https://azure.microsoft.com/) account with an active subscription and [Azure Database for PostgreSQL - Flexible Server instance](https://learn.microsoft.com/en-gb/azure/postgresql/flexible-server/overview).
```

docs/external-db.md
```diff
@@ -0,0 +1,159 @@
...
+- Download Microsoft RSA Root Certificate Authority 2017 and DigiCert Global Root CA certificates from the URIs provided [here](https://learn.microsoft.com/en-gb/azure/postgresql/flexible-server/concepts-networking-ssl-tls#downloading-root-ca-certificates-and-updating-application-clients-in-certificate-pinning-scenarios)
+
+**NOTE:**  Azure Database for PostgreSQL flexible server **enforces** connecting your client applications using Transport Layer Security (TLS)
+
+- Convert .crt files you downloaded to .pem format as [suggested](https://learn.microsoft.com/en-gb/azure/postgresql/flexible-server/concepts-networking-ssl-tls#downloading-root-ca-certificates-and-updating-application-clients-in-certificate-pinning-scenarios) using
+
+````
+openssl x509 -in DigiCertGlobalRootCA.crt -out DigiCertGlobalRootCA.crt.pem -outform PEM
+
+openssl x509 -in Microsoft ECC Root Certificate Authority 2017.crt -out Microsoft ECC Root Certificate Authority 2017.crt.pem -outform PEM
```
rm3l, line 152:
```suggestion
openssl x509 -in "Microsoft ECC Root Certificate Authority 2017.crt" -out "Microsoft ECC Root Certificate Authority 2017.crt.pem" -outform PEM
```

docs/external-db.md
```diff
@@ -0,0 +1,159 @@
...
+````
+
+- Combine them according to this [suggestion](https://learn.microsoft.com/en-gb/azure/postgresql/flexible-server/how-to-update-client-certificates-java#updating-root-ca-certificates-for-other-clients-for-certificate-pinning-scenarios), like:
+
+````
+cat DigiCertGlobalRootCA.crt.pem Microsoft ECC Root Certificate Authority 2017.crt.pem > postgres-crt.pem
```
rm3l, line 158:
```suggestion
cat DigiCertGlobalRootCA.crt.pem "Microsoft ECC Root Certificate Authority 2017.crt.pem" > postgres-crt.pem
```

rm3l (changes requested):

docs/external-db.md
gazarenkov: @kim-tsao Well, I have instruction for Helm Chart too, just thinking how to organize it not duplicating Postgre SQL types part (same for both). Any suggestions?

docs/external-db.md
```diff
@@ -0,0 +1,159 @@
...
+metadata:
+ name: <crt-secret>
+type: Opaque
+stringData:
+ postgres-crt.pem: |-
+   -----BEGIN CERTIFICATE-----
+   MIIFqDCCA5CgAwIBAgIQHtOXCV/YtLNHcB6qvn9FszANBgkqhkiG9w0BAQwFADBl
+   ... 
+````
+
```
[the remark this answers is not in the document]
gazarenkov: It is individual, per PGSQL server, you can see it below

docs/external-db.md
gazarenkov: Yes, it would achieve the same result and talking to users I realised they tend to use the extra-secrets way as a generic approach to inject env variables (not necessarily for app-config), that's why I decided to promote this way.

docs/external-db.md
```diff
@@ -0,0 +1,159 @@
...
+Backstage hosts the data in the [PostgreSQL database](https://backstage.io/docs/getting-started/config/database/).
+By default, Operator creates and manages local instance of PostgreSQL in the same as Backstage deployment namespace but it also allows to switch this off and configure external instance instead.
+Usually, external connection requires more security, so, this instruction includes steps to configure SSL/TLS.
+
+### Configure your external PostgreSQL instance
+As a prerequisite you have to know:
+- **db-host** - your PostgreSQL instance DNS or IP address 
+- **db-port** - your PostgreSQL instance port number (usually 5432)
+- **username** - to connect to your PostgreSQL instance
```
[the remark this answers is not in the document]
gazarenkov: added

docs/external-db.md
```diff
@@ -0,0 +1,161 @@
...
+
+- disable creating local PostgreSQL instance with **spec.database.enableLocalDb: false**
+- add **<crt-secret>** to **spec.application.extraFiles.secrets**, so, as for example below **postgres-crt.pem** file will be mounted to Backstage container at **spec.application.extraFiles.mountPath** directory:   
+- add **<cred-secret>** to **spec.application.extraEnvs.secrets**, so all the data's entries will be injected to Backstage container as environment variables.
+
+**NOTE:** environment variables listed in **<cred-secret>** file work with default Operator configuration. If it is changed on default or raw configuration, you have to re-configure it accordingly.
+
+````
+cat <<EOF | kubectl -n <your-namespace> create -f -
+apiVersion: v1
```
rm3l, line 67:
```suggestion
apiVersion: rhdh.redhat.com/v1alpha1
```

docs/external-db.md
```diff
@@ -0,0 +1,161 @@
...
+kind: Secret
```
rm3l, line 68:
```suggestion
kind: Backstage
```

rm3l (changes requested):

docs/external-db.md
```diff
@@ -0,0 +1,159 @@
...
+metadata:
+ name: <cred-secret-name>
+type: Opaque
+stringData:
+ POSTGRES_PASSWORD: <password>
+ POSTGRES_PORT: <db-port>
+ POSTGRES_USER: <username>
+ POSTGRES_HOST: <db-host>
+ PGSSLMODE: require #  for TLS connection
```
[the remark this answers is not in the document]
rm3l: @gazarenkov Do your instructions work with this stricter mode?
If they do, I think we should recommend this since it provides both eavesdropping and MITM protection. See https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-PROTECTION
cc @kim-tsao

docs/external-db.md
```diff
@@ -0,0 +1,161 @@
...
+Usually, external connection requires more security, so, this instruction includes steps to configure SSL/TLS.
+
+### Configure your external PostgreSQL instance
+As a prerequisite, you have to know:
+- **db-host** - your PostgreSQL instance DNS or IP address 
+- **db-port** - your PostgreSQL instance port number (usually 5432)
+- **username** - to connect to your PostgreSQL instance
+- **password** - to connect to your PostgreSQL instance
+
+**NOTE:** By default, Backstage uses databases for each plugin and automatically creates them if none are found, so in addition to PSQL Database level privileges, the user may need Create Database privilege.
```
rm3l, line 14: Per https://backstage.io/docs/tutorials/configuring-plugin-databases/#check-your-databases, permission to list or show databases might also be required. Could we link to this webpage?

```suggestion
**NOTE:** By default, Backstage uses databases for each plugin and automatically creates them if none are found, so in addition to PSQL Database level privileges, the user may need Create Database privilege.  More details in [Check Your Databases](https://backstage.io/docs/tutorials/configuring-plugin-databases/#check-your-databases).
```

docs/external-db.md
```diff
@@ -0,0 +1,161 @@
...
+metadata:
+ name: <cred-secret-name>
+type: Opaque
+stringData:
+ POSTGRES_PASSWORD: <password>
+ POSTGRES_PORT: <db-port>
+ POSTGRES_USER: <username>
+ POSTGRES_HOST: <db-host>
+ PGSSLMODE: require #  for TLS connection
+ NODE_EXTRA_CA_CERTS: <abs-path-to-pem-file> # for TLS connection, e.g. /opt/app-root/src/postgres-crt.pem
```
rm3l, line 36: Per https://backstage.io/docs/tutorials/switching-sqlite-postgres/#add-postgresql-configuration, there is a `backend.database.connection.ssl` object in the app-config, where we can configure the `node-postgres` client with things like a client certificate file for example (besides a CA cert file) - see https://node-postgres.com/features/ssl.
Did you test this path instead?
Maybe I've missed something, but I am not sure specifying a client certificate is do-able via the `NODE_EXTRA_CA_CERTS` env var (it covers CA certs, I guess).

rm3l: nvm. Per https://node-postgres.com/features/connecting, it seems the `node-postgres` client honors the same [environment variables](https://www.postgresql.org/docs/current/libpq-envars.html#LIBPQ-ENVARS) as `psql` to connect to a PostgreSQL server. So in the case of a client cert, I guess we can pass a `PGSSLCERT` env var.
But maybe it is worth mentioning somewhere that those environment variables should work?

docs/external-db.md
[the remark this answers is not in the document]
gazarenkov: it is the same in this case, "verify-full" implied if you configure certificate.
(I even think this env var is not mandatory in this case, you can try it yourself ;) )

docs/external-db.md
gazarenkov: I seens, yes :) but it did not work for me, show me please if it works for you.

docs/external-db.md
gazarenkov: Not sure I get your "more user friendly" point, of course there are more than one way to do it but IMO it is more convenient to copy/paste/change one command than do the same but with one file AND one command... matter of taste :)

rm3l: I mean, a command like this:
```shell
kubectl create secret generic <crt-secret> --from-file=postgres-crt.pem=/path/to/pem/file
```

is more convenient than the current `cat <<EOF | kubectl -n <your-namespace> create -f -` command, which requires you to paste the content of your cert file in the `stringData` field (and make sure to indent correctly)

But anyway, I'm fine with what is here as well.

rm3l (approved):

Issue #355, opened 2024-05-07
Potential breaking change between 1.1.x and main (1.2) for the default Route URL naming pattern


rm3l:
/kind bug

## What versions of software are you using?

- Operator version: latest commit on main: c889b0e52bca79a7d5dfd5424165b735dc5103b7

- OpenShift Local:
```
CRC version: 2.35.0+3956e8
OpenShift version: 4.15.10
Podman version: 4.4.4
```

## How did you run the operator exactly?

1. Started a fresh CRC cluster (no previous Backstage applications)
2. Took an existing CR with an app-config ConfigMap that was known to be working with 1.1:

<details>
<summary>Details</summary>

```yaml
apiVersion: rhdh.redhat.com/v1alpha1
kind: Backstage
metadata:
  name: my-rhdh
spec:
  application:
    appConfig:
      configMaps:
        - name: app-config-rhdh

---
apiVersion: v1
kind: ConfigMap
metadata:
  name: app-config-rhdh
data:
  "app-config-rhdh-app.yaml": |
    app:
      baseUrl: https://backstage-my-rhdh-my-ns.apps-crc.testing

  "app-config-rhdh-backend.yaml": |
    backend:
      baseUrl: https://backstage-my-rhdh-my-ns.apps-crc.testing
      csp:
        connect-src:
          - "'self'"
          - 'http:'
          - 'https:'
      cors:
        origin: https://backstage-my-rhdh-my-ns.apps-crc.testing
        methods: [GET, HEAD, PATCH, POST, PUT, DELETE]
        credentials: true
      auth:
        dangerouslyDisableDefaultAuthPolicy: true
        keys:
          - secret: "xxx"
```

</details>

3. Started the operator from the main branch and deploy the manifest above
5. Wait until the Backstage application pod is ready.

## Actual behavior

The documentation in 1.1 depicts the URL naming format (used for `backend.baseUrl`, `app.baseUrl` and `backend.cors.origin`) as: `https://backstage-<CUSTOM_RESOURCE_NAME>-<NAMESPACE_NAME>.<OPENSHIFT_INGRESS_DOMAIN>`.
But if we look at the Route created by the operator on main, it now has the following naming format: `https://<CUSTOM_RESOURCE_NAME>-backstage-<NAMESPACE_NAME>.<OPENSHIFT_INGRESS_DOMAIN>`

## Expected behavior

Changing the URL format will break users' configurations, as they will need to change them manually to make it work properly with the Route created by the Operator.

[...]

PR #369, opened 2024-05-29
fix: Make sure to use backward-compatible names for resources created by the operator [RHIDP-2432]

rm3l:
## Description

This fixes a critical issue where, when upgrading existing resources using an operator upgraded from 1.1 to 1.2 (`main` branch), the operator would create and use a brand new database StatefulSet. This would cause the upgraded Backstage application to boot using an empty database, rather than starting with the existing database (which could already contain some data).

It also fixes a more general issue where resource naming was not backward-compatible, which would create for example 2 different ConfigMaps after an upgrade: one named `backstage-dynamic-plugins-<cr>` (created by the operator on the `main` branch) besides an already existing one named `<cr>-dynamic-plugins` (created in 1.1), which could create some confusion when looking at the resources in the namespace.

- Fixes https://issues.redhat.com/browse/RHIDP-2432

## PR acceptance criteria

- [x] Tests
- [ ] Documentation

See the repro steps depicted in https://issues.redhat.com/browse/RHIDP-2432 and run the tests with the following command:
```
make test
make integration-test \
  ARGS=-focus='creates runtime object using raw configuration then updates StatefulSet to replace some immutable fields' \
  USE_EXISTING_CLUSTER=true
```

### DIFF
--- a/controllers/backstage_controller.go
+++ b/controllers/backstage_controller.go
@@ -181,11 +181,22 @@ func (r *BackstageReconciler) applyObjects(ctx context.Context, objects []model.
 		}
 
 		if err := r.patchObject(ctx, baseObject, obj); err != nil {
+			_, isStatefulSet := obj.Object().(*appsv1.StatefulSet)
+			if isStatefulSet && errors.IsInvalid(err) && errors.HasStatusCause(err, metav1.CauseTypeForbidden) {
+				// Some resources like StatefulSets allow patching a limited set of fields.
+				// That's why we are trying to delete them first, taking care of orphaning the dependents so that resources like PVCs can be retained.
+				// They will be recreated at the next reconciliation.
+				if err = r.Delete(ctx, baseObject, client.PropagationPolicy(metav1.DeletePropagationOrphan)); err != nil {
+					return fmt.Errorf("failed to delete object %s so it can be recreated: %w", obj.Object(), err)
+				}
+				lg.V(1).Info("delete object ", objDispName(obj), obj.Object().GetName())
+				continue
+			}
+
 			return fmt.Errorf("failed to patch object %s: %w", obj.Object(), err)
 		}
 
 		lg.V(1).Info("patch object ", objDispName(obj), obj.Object().GetName())
-
 	}
 	return nil
 }
--- a/integration_tests/db_test.go
+++ b/integration_tests/db_test.go
@@ -16,19 +16,22 @@ package integration_tests
 
 import (
 	"context"
+	"fmt"
 	"time"
 
-	"k8s.io/apimachinery/pkg/api/errors"
+	"redhat-developer/red-hat-developer-hub-operator/pkg/model"
 
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 
 	corev1 "k8s.io/api/core/v1"
 
 	appsv1 "k8s.io/api/apps/v1"
 
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-
-	"redhat-developer/red-hat-developer-hub-operator/pkg/model"
 
 	bsv1alpha1 "redhat-developer/red-hat-developer-hub-operator/api/v1alpha1"
 
@@ -63,10 +66,10 @@ var _ = When("create backstage with CR configured", func() {
 			err := k8sClient.Get(ctx, types.NamespacedName{Namespace: ns, Name: model.DbStatefulSetName(backstageName)}, &appsv1.StatefulSet{})
 			g.Expect(err).To(Not(HaveOccurred()))
 
-			err = k8sClient.Get(ctx, types.NamespacedName{Namespace: ns, Name: model.DbStatefulSetName(backstageName)}, &corev1.Service{})
+			err = k8sClient.Get(ctx, types.NamespacedName{Namespace: ns, Name: model.DbServiceName(backstageName)}, &corev1.Service{})
 			g.Expect(err).To(Not(HaveOccurred()))
 
-			err = k8sClient.Get(ctx, types.NamespacedName{Namespace: ns, Name: model.DbStatefulSetName(backstageName)}, &corev1.Secret{})
+			err = k8sClient.Get(ctx, types.NamespacedName{Namespace: ns, Name: model.DbSecretDefaultName(backstageName)}, &corev1.Secret{})
 			g.Expect(err).To(Not(HaveOccurred()))
 
 		}, time.Minute, time.Second).Should(Succeed())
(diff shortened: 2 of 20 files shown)


integration_tests/db_test.go
```diff
@@ -143,4 +146,234 @@ var _ = When("create backstage with CR configured", func() {
...
+								},
+							},
+						},
+					},
+				},
+			})
+			Expect(err).ShouldNot(HaveOccurred())
+		})
+
+		It("should reuse existing DB resources when reconciling", func() {
```
gazarenkov, line 325: Not sure what do we test with such a complex procedure...
It looks like that K8s correctly uses object type/name as a key and does not recreate it? Is it the point?

pkg/utils/utils.go
```diff
@@ -51,7 +51,12 @@ func GenerateLabel(labels *map[string]string, name string, value string) {
 }
 
 // GenerateRuntimeObjectName generates name using BackstageCR name and objectType which is ConfigObject Key without '.yaml' (like 'deployment')
-func GenerateRuntimeObjectName(backstageCRName string, objectType string) string {
+func GenerateRuntimeObjectName(backstageCRName string, objectType string, crNameAsPrefix bool) string {
```
gazarenkov, line 54: I would really like to have it unified, as discussed and not to make us remember what is the pattern for each object type.  I do not think it impacts on functionality, the only drawback is that we leave old objects which can be solved by:
- upgrade documentation (recommending to remove it manually when upgrade from 1.1)
- additional ckeck/remove  (but I'd avoid it).

gazarenkov (changes requested):
I would not like to have different name patterns for generated objects forever, let's unify it sooner than later.

controllers/backstage_controller.go
```diff
@@ -181,7 +181,27 @@ func (r *BackstageReconciler) applyObjects(ctx context.Context, objects []model.
 		}
 
 		if err := r.patchObject(ctx, baseObject, obj); err != nil {
-			return fmt.Errorf("failed to patch object %s: %w", obj.Object(), err)
```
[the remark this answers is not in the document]
rm3l: I find it more readable as it allows for a better understanding of the intent and exit conditions rather than long statements.
But also, as generally practiced, trying to keep the normal code path at a minimal indentation.
Anyway, I've tried to simplify this - see [`ec43e69` (#369)](https://github.com/janus-idp/operator/pull/369/commits/ec43e699618841fa0e68d42dd8523e6966a487a0)

pkg/utils/utils.go
rm3l: > I would really like to have it unified, as discussed and not to make us remember what is the pattern for each object type. I do not think it impacts on functionality
>
Well, in some cases, it impacted functionally. For example, I had to revert the DB Secret name to make sure a new Secret was not being created. Otherwise, connection to the existing DB would just fail.
I agree with you (at least in the medium/long run). But at the same time, found it really confusing to have both old and new objects in the cluster (without anything conveying which ones are deprecated).
I remember from a discussion someone edited the `<cr_name>-dynamic-plugins` ConfigMap (name in 1.1) to update their list of dynamic plugins. How would we handle this now that the name has changed to `backstage-dynamic-plugins-<cr_name>` in 1.2?

This is not ideal, but for the upcoming 1.2, I think it is safer to keep the existing naming to minimize the risk of additional bugs.

image

> the only drawback is that we leave old objects which can be solved by:
> 
> * upgrade documentation (recommending to remove it manually when upgrade from 1.1)
> * additional ckeck/remove  (but I'd avoid it).

The problem I foresee is with the Automatic Upgrade strategy with OLM, which, in my understanding, is much appreciated by users. IMO, we cannot claim backward compatibility and expect CR users to perform manual cleanup operations after an upgrade of the application (potentially automatic).
Or maybe a K8s Job (created by the operator) that does this migration automatically might make sense.
But that would be a separate issue.

integration_tests/db_test.go
gazarenkov: Sorry, still a bit confusing to me :) 

It sounds like you want to check 2 things:
1. the names of objects created by operator are expected
2. statefulset "patching" in a case when we try to change immutable fields (serviceName?) works as expected (delete/create)

If that's the case I believe have 2 simpler tests. First as unit test (just compare your model with predictable object names), second as integration test where you create CR and then update it (using raw config apparently) with changed SS. 
Does it make sense?

controllers/backstage_controller.go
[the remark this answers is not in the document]
gazarenkov: Well, TBH I am not sure that this kind of "patching" is 100% correct (indeed there is a reason to keep some SS's fields immutable?) so I 'd like this patch looks as "isolated" as possible and would not look like the main flow.
Apparently we can not avoid it for upgrading but I would consider it as a rather temporary TBH.

pkg/utils/utils.go
gazarenkov: To make it clear

What is important to me is that the names of objects generated by Operator have unified pattern.
I.e. (as discussed) backstage-* or backstage-psql-* shape.
It is important for me that all of auto-generated items looks really as auto-generated, it also will simplify support.
And I think the sooner we do it the better (i.e. if we do not do it now it is a big chance it will live forever).

For your notes:

I think DB secret should not be affected, as I can see 1.1 name is secret/backstage-psql-secret-bs1 which fits the pattern, no?

For dynamic-plugins, I am not sure I understand the problem.
If you mean user change operator-config's configmap or secret - I do not think it will work anyway since it will be changed to desired state (from  operator-config). Or you meant something else? 

So far I can not see any backward compatibility problem.
I do not think existence of "old" objects can be taken as incompatible thing. 
Yes, we need to mention it in the Release Notes and recommend to clean it (I would rather not spend time for dedicated tooling for it).

controllers/backstage_controller.go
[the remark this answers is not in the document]
rm3l: I took a quick look at how more specialized DB Operators would reconcile StatefulSets, and they appear to handle this the same way. See:
- https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/resources.go#L198-L238
- https://github.com/CrunchyData/postgres-operator/blob/master/internal/controller/postgrescluster/pgadmin.go#L330-L351
- https://github.com/percona/percona-postgresql-operator/blob/main/internal/controller/postgrescluster/pgadmin.go#L332-L353

So, unless I've missed something, it seems to me that would be part of the "normal" code path.

pkg/utils/utils.go
rm3l: What pushes me to stick to the previous naming in 1.1 is because this is the second issue caused by this naming change (first one was https://github.com/janus-idp/operator/issues/355). And I wanted to prevent further issues due to this.
But alright, let's go ahead with your suggestions. I'll update the release notes text on the JIRA issue accordingly.
Thanks.

rm3l:
@gazarenkov Thanks for the review comments. Apart from the one about the integration tests (which I think make sense as a way to guarantee the behavior and prevent further regression on this), I think it should be good. PTAL and approve if it is okay for you.

controllers/backstage_controller.go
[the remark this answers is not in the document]
gazarenkov: OK
My proposal was mostly about to move this logic from to patchObject function since it is patch specific, nothing critically important.

integration_tests/db_test.go
gazarenkov: I think the tests showl be as focused as possible. It is still not clear to what does this test check. 

The way I see it is that you try to cover 2 cases: 
- the names of objects we create are expected. I.e:
backstage-<cr-name>,  not <cr-name>-backstage
backstage-psql-<cr-name>,  not backstage-db-<cr-name>
etc.
This one can be covered by simple unit test IMO?
- statefulset "patching" in a case when we try to change immutable fields (serviceName?) works as expected (delete/create), i.e:
the (integration) test can create CR with serviceName expected by 1.1 using raw configuration 
then it can update CR with default serviceName and validate it.
WDYT?

gazarenkov (commented):
I'd like to see more targeted tests as mentioned in the comments, otherwise I doubt we'll be able to figure out what it was testing any time soon.

gazarenkov (commented):
I'd like to see more targeted tests as mentioned in the comments, otherwise I doubt we'll be able to figure out what it was testing any time soon.

integration_tests/db_test.go
rm3l: Okay, I've tried to change the tests with this trade-off. See [`4758c86` (#369)](https://github.com/janus-idp/operator/pull/369/commits/4758c86e5758f2e294db3c7d8aa0ea31dd6cc8bd)

integration_tests/default-config_test.go
```diff
@@ -156,4 +160,64 @@ var _ = When("create default backstage", func() {
...
+			})
+			Expect(err).To(Not(HaveOccurred()))
+		}
+
+		Eventually(func(g Gomega) {
+			By("replacing StatefulSet")
+			dbStatefulSet := &appsv1.StatefulSet{}
+			err = k8sClient.Get(ctx, types.NamespacedName{Namespace: ns, Name: fmt.Sprintf("backstage-psql-%s", backstageName)}, dbStatefulSet)
+			g.Expect(err).ShouldNot(HaveOccurred())
+			g.Expect(dbStatefulSet.Spec.ServiceName).To(Equal(fmt.Sprintf("backstage-psql-%s-hl", backstageName)))
```
gazarenkov, line 219: it will be backstage-psql-<backstageName> after https://github.com/janus-idp/operator/pull/371

integration_tests/default-config_test.go
```diff
@@ -156,4 +160,64 @@ var _ = When("create default backstage", func() {
...
+			ss := &appsv1.StatefulSet{}
+			name := fmt.Sprintf("backstage-psql-%s", backstageName)
+			err = k8sClient.Get(ctx, types.NamespacedName{Namespace: ns, Name: name}, ss)
+			g.Expect(err).ShouldNot(HaveOccurred())
+			g.Expect(ss.Spec.Template.Spec.Containers).To(HaveLen(1))
+			g.Expect(ss.Spec.Template.Spec.Containers[0].Image).To(Equal("busybox"))
+			g.Expect(ss.Spec.ServiceName).To(Equal("backstage-psql-db-statefulset-hl"))
+		}, time.Minute, time.Second).Should(Succeed())
+
+		By("updating CR with new DB raw config")
```
gazarenkov, line 195: Up to you, but, why do you need to update raw config?
IMO it is enough to just update Backstage with default config (especially after https://github.com/janus-idp/operator/pull/371 ) no?

rm3l: I wanted to check that something had actually changed to ensure the StatefulSet replacement. So it should be a field that is not overwritten upon the first reconciliation.
But with #371 in mind, I'll change that to another immutable field and update Backstage to the default config. Thanks.

rm3l: Updated in [`0f194b2` (#369)](https://github.com/janus-idp/operator/pull/369/commits/0f194b28ebcb5f9374d96c8e3ba781484c23fd5b). PTAL.

rm3l:
/cherry-pick 1.2.x

But putting on hold until https://github.com/janus-idp/operator/pull/376 is merged, to avoid the change to the CSV.

/hold

gazarenkov (approved):

rm3l:
/hold cancel

Rebased onto `main`, now that https://github.com/janus-idp/operator/pull/376 has been merged.

gazarenkov (approved):

PR #372, opened 2024-05-30
added telemetry opt-in banner

04kash:
## Description
Added the telemetry opt-in banner to the following files: `.rhdh/bundle/manifests/rhdh-operator.csv.yaml`, `README.md`, and `config/manifests/bases/backstage-operator.clusterserviceversion.yaml`

- Fixes https://issues.redhat.com/browse/RHIDP-1907

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation


README.md
```diff
@@ -59,6 +59,9 @@ Output:
 - [Developer Guide](docs/developer.md)
 - [Operator Design](docs/developer.md)
 
+## Telemetry data collection
+
+The telemetry data collection feature is enabled by default. Red Hat Developer Hub sends telemetry data to Red Hat by using the `backstage-plugin-analytics-provider-segment` plugin. To disable this and to learn what data is being collected, see https://access.redhat.com/documentation/en-us/red_hat_developer_hub/1.2/html-single/administration_guide_for_red_hat_developer_hub/index#assembly-rhdh-telemetry_admin-rhdh
```
nickboldt, line 64: Since this repo may at some point become the upstream Backstage operator, I'm not sure we want to mention RHDH in the root readme. WDYT @rm3l  @gazarenkov ? Are we still planning to keep this project "whitelabelled" to eventually move upstream?

.rhdh/bundle/manifests/rhdh-operator.csv.yaml
```diff
@@ -62,6 +62,8 @@ spec:
 
     System Architects can benefit by implementing a tailored platform with a complementary suite of verified and curated tools and components needed for operations teams to support developers—within a centralized, consistent location. Development teams can experience increased productivity, fewer development team obstacles, and simplified governance of technology choices with self-service and guardrails.
 
+    The telemetry data collection feature is enabled by default. Red Hat Developer Hub sends telemetry data to Red Hat by using the `backstage-plugin-analytics-provider-segment` plugin. To disable this and to learn what data is being collected, see https://access.redhat.com/documentation/en-us/red_hat_developer_hub/1.2/html-single/administration_guide_for_red_hat_developer_hub/index#assembly-rhdh-telemetry_admin-rhdh
```
nickboldt, line 65: since the version is hardcoded here, we will need to make sure that this version is updated at intervals when we branch for 1.2.x and main branch moves to 1.3. See https://gitlab.cee.redhat.com/rhidp/rhdh/-/blob/rhdh-1-rhel-9/build/scripts/tagRelease.sh#L317-322

config/manifests/bases/backstage-operator.clusterserviceversion.yaml
```diff
@@ -17,7 +17,11 @@ spec:
       kind: Backstage
       name: backstages.rhdh.redhat.com
       version: v1alpha1
-  description: Operator to deploy Backstage on Kubernetes
+  description: |
+    Operator to deploy Backstage on Kubernetes
+
+    The telemetry data collection feature is enabled by default. Red Hat Developer Hub sends telemetry data to Red Hat by using the `backstage-plugin-analytics-provider-segment` plugin. To disable this and to learn what data is being collected, see https://access.redhat.com/documentation/en-us/red_hat_developer_hub/1.2/html-single/administration_guide_for_red_hat_developer_hub/index#assembly-rhdh-telemetry_admin-rhdh
```
nickboldt, line 23: same comment as above -- see upstream doc in https://github.com/janus-idp/backstage-showcase/pull/1028/files#diff-8b34298b8105f19a866e1072c43ff5fa2675f4b29a8511a74e6152c88a171718R9-R29 ... no mention of RHDH in the upstream.

README.md
gazarenkov: I would not like we have direct link to RedHat documentation here.
Let's host information of how to disable/enable telemetry plugin here, on Operator repo (see comment below).

config/manifests/bases/backstage-operator.clusterserviceversion.yaml
gazarenkov: I would prefer to move operator part from showcase document to operator repo (as TELEMETRY.md or so) and maintain it here. The reason is that we are evolving and I'd like to make sure the info is updated.
The message should be like - default Backstage operator configuration (which is RHDH) contains `backstage-plugin-analytics-provider-segment` dynamic plugin which sends telemetry data to Red Hat. To disable this and to learn what data is being collected <TELEMETRY.md>

gazarenkov (changes requested):

README.md
rm3l, line 64: Good point. How about something like this? As commented out below, I think we should link to the Showcase docs instead.

```suggestion
The telemetry data collection feature is enabled by default. In the default configuration of the operator, telemetry data will be sent to Red Hat by using the `backstage-plugin-analytics-provider-segment` plugin. To disable this and to learn what data is being collected, see https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/getting-started.md#telemetry-collection
```

config/manifests/bases/backstage-operator.clusterserviceversion.yaml
```diff
@@ -17,7 +17,10 @@ spec:
...
```
rm3l, line 23: Same suggestion as above:

```suggestion
    The telemetry data collection feature is enabled by default. In the default configuration of the operator, telemetry data will be sent to Red Hat by using the `backstage-plugin-analytics-provider-segment` plugin. To disable this and to learn what data is being collected, see https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/getting-started.md#telemetry-collection
```

> I would prefer to move operator part from showcase document to operator repo (as TELEMETRY.md or so) and maintain it here. The reason is that we are evolving and I'd like to make sure the info is updated.

Even if this [doc](https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/getting-started.md#disable-telemetry) mentions the operator, I read it as more about disabling telemetry in the Showcase app, not telemetry from the operator itself.
So I think it makes sense to be in the Showcase docs, as the operator may need to support multiple versions of the Showcase app (and the way telemetry is configured in the app might be different between 2 versions of the app).
Maybe we will need to also send telemetry data from the operator itself later, in which case a `TELEMETRY.md` file here would make sense, I guess.

WDYT?

.rhdh/bundle/manifests/rhdh-operator.csv.yaml
rm3l: Good point. Just curious, is it possible to link to a kind of "latest" link of the doc?

rm3l:
/hold

https://access.redhat.com/documentation/en-us/red_hat_developer_hub/1.2/html-single/administration_guide_for_red_hat_developer_hub/index#assembly-rhdh-telemetry_admin-rhdh currently returns a 404.
Should we wait until confirmation that the URL is correct?
(Feel free to remove the hold if you think otherwise).

config/manifests/bases/backstage-operator.clusterserviceversion.yaml
gazarenkov: Yes, it is Showcase's configuration but user configures it using Operator (in this case)
The way Operator makes this configuration is Operator specific and may be changed or new options may be introduced in the future.
So, IMO, it will be simpler to maintain it on Operator side.

rm3l: > Yes, it is Showcase's configuration but user configures it using Operator (in this case) The way Operator makes this configuration is Operator specific and may be changed or new options may be introduced in the future. So, IMO, it will be simpler to maintain it on Operator side.
> 
> In other words: if we keep Operator specific configuration in Showcase repo, who will maintain it? :)

I understand your point, but this can indeed evolve in 2 ways:
- either the Showcase changes the way telemetry is configured. In this case, I think it makes sense to have the related doc in the Showcase repo.
- or the Operator changes the way dynamic plugins are defined in the Custom Resource (which can have an impact on how telemetry is configured). In this case, I agree with you that it makes sense to have the doc in the Operator repo.

Right now, I see it as documenting the specific Showcase telemetry using all possible installation methods on the same doc. And the Operator has a more general section about configuring dynamic plugins: https://github.com/janus-idp/operator/blob/main/.rhdh/docs/openshift.adoc#configuring-dynamic-plugins-with-the-operator

But your question is more general to me and should be handled outside of this PR.
As this doc was added in https://github.com/janus-idp/backstage-showcase/pull/1028 (https://issues.redhat.com/browse/RHIDP-1311), let's file a separate issue for this if you think the doc content should be reorganized.

gazarenkov: My question is not that general :), it is as simple as: who and how will maintain actuality of Operator configuration for this case?
As for your examples:
- if something changed on Showcase, Operator will have to react and change something too (including doc apparently)
- but not opposite, i.e. if something changed on Operator, most probably Showcase need not to react.
That's why Showcase part of document should rather contain a reference to Operator/Helm specific and not Operator/Helm specific information itself.
Does it make sense?

rm3l: > My question is not that general :), it is as simple as: who and how will maintain actuality of Operator configuration for this case?
>
In this case, the person who made the change will be responsible for ensuring the Showcase docs is also updated.
But I said "general" because I saw this pattern in almost all Showcase docs (like the docs for adding monitoring, logging, corporate proxy - see https://github.com/janus-idp/backstage-showcase/tree/main/showcase-docs).

If tomorrow we decide to make telemetry opt-in by default in the Showcase app, your suggestion means we would have to create 3 different PRs (one in Showcase, and one per installation method), right? Rather than a single PR in the Showcase repo?

config/manifests/bases/backstage-operator.clusterserviceversion.yaml
kadel: The operator itself doesn't collect any telemetry; it is RHDH/JanusShowcase that does that. Keeping separate TELEMETRY.md in this repo might be confusing, and it would definitely create a duplication as this information is already managed in [Janus Showcase dosc](https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/getting-started.md#telemetry-collection) and RHDH docs

kadel: We can link to the upstream  docs (https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/getting-started.md#telemetry-collection) here, but we need to make sure that the during RHDH build it is updated to the downstream docs.


We can also reword it so it doesn't mention the product name to make it easier to update.

> The telemetry data collection feature is enabled by default. The default configuration uses image with `backstage-plugin-analytics-provider-segment` plugin enabled and configured. To disable this and to learn what data is being collected, see ... 

link for upstream (this repo): 
https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/getting-started.md#telemetry-collection
link for downstream (RHDH build):
https://access.redhat.com/documentation/en-us/red_hat_developer_hub/1.2/html-single/administration_guide_for_red_hat_developer_hub/index#assembly-rhdh-telemetry_admin-rhdh

.rhdh/bundle/manifests/rhdh-operator.csv.yaml
nickboldt: nope, there's no "latest" deep doc linking. https://access.redhat.com/documentation/en-us/red_hat_developer_hub/ will bounce to the latest release https://access.redhat.com/documentation/en-us/red_hat_developer_hub/1.1 but that won't get you to the right spot :(

README.md
nickboldt: +1.

bundle/manifests/backstage-operator.clusterserviceversion.yaml
```diff
@@ -37,7 +37,10 @@ spec:
       kind: Backstage
       name: backstages.rhdh.redhat.com
       version: v1alpha1
-  description: Operator to deploy Backstage on Kubernetes
+  description: |
+    Operator to deploy Backstage on Kubernetes
+
+    The telemetry data collection feature is enabled by default. Red Hat Developer Hub sends telemetry data to Red Hat by using the `backstage-plugin-analytics-provider-segment` plugin. To disable this and to learn what data is being collected, see https://access.redhat.com/documentation/en-us/red_hat_developer_hub/1.2/html-single/administration_guide_for_red_hat_developer_hub/index#assembly-rhdh-telemetry_admin-rhdh
```
nickboldt, line 43: this will need to be changed similar ot https://github.com/janus-idp/operator/pull/372/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R64 so we point to showcase, not RHDH.

config/manifests/bases/backstage-operator.clusterserviceversion.yaml
04kash: I've reworded it and also updated the link to the one mentioned for upstream

.rhdh/bundle/manifests/rhdh-operator.csv.yaml
```diff
@@ -62,6 +62,8 @@ spec:
 
     System Architects can benefit by implementing a tailored platform with a complementary suite of verified and curated tools and components needed for operations teams to support developers—within a centralized, consistent location. Development teams can experience increased productivity, fewer development team obstacles, and simplified governance of technology choices with self-service and guardrails.
 
+    The telemetry data collection feature is enabled by default. The default configuration uses image with backstage-plugin-analytics-provider-segment plugin enabled and configured. To disable this and to learn what data is being collected, see https://github.com/janus-idp/backstage-showcase/blob/main/showcase-docs/getting-started.md#telemetry-collection
```
nickboldt, line 65: this one should point to RHDH since it's the RHDH operator. 
```suggestion
    The telemetry data collection feature is enabled by default. Red Hat Developer Hub sends telemetry data to Red Hat by using the `backstage-plugin-analytics-provider-segment` plugin. To disable this and to learn what data is being collected, see https://access.redhat.com/documentation/en-us/red_hat_developer_hub/1.2/html-single/administration_guide_for_red_hat_developer_hub/index#assembly-rhdh-telemetry_admin-rhdh
```

.rhdh/bundle/manifests/rhdh-operator.csv.yaml
nickboldt: @04kash might want to add something for this into your MR https://gitlab.cee.redhat.com/rhidp/rhdh/-/merge_requests/21 (or else a second MR)

nickboldt (changes requested):

nickboldt (approved):

PR #376, opened 2024-06-04
Fix version number in `skipRange` and `name` fields in base CSV

rm3l:
## Description
This fixes the `skipRange` field in the base CSV on the `main` branch.
After rebasing https://github.com/janus-idp/operator/pull/369 onto `main`, I noticed a change to the generated bundle, which should not have occurred there.

This is a follow-up to https://github.com/janus-idp/operator/pull/375

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation


rm3l:
> It is generated file, I believe it uses kustomization.yaml

I think Kustomize picks `config/manifests/bases/backstage-operator.clusterserviceversion.yaml` as the base: https://github.com/janus-idp/operator/blob/main/config/manifests/kustomization.yaml#L4 

So I don't think it is generated, but it serves as the base CSV used to generate the one in `bundle/manifests`

gazarenkov (approved):

Issue #129, opened 2024-01-10
Configuration documentation


gazarenkov:
/docs/configuration.md

- [ ] Document how to enable/disable auto-sync upon config changes (https://github.com/janus-idp/operator/pull/351#discussion_r1593693626)

gazarenkov:
NOTE: need to add information about dynamic-plugins file name specific as mentioned in https://github.com/janus-idp/operator/pull/381

nickboldt:
As part of the migration from janus-idp to redhat-developer in https://issues.redhat.com/browse/RHIDP-1021, this will be tracked in https://issues.redhat.com/browse/RHIDP-1658

PR #381, opened 2024-06-05
Notify about dynamic plugin's file name constraints.

gazarenkov:
## Description

The problem is that dynamic plugins are silently not loaded in a case if the configured filename (key) in the configMap is not "dynamic-plugins.yaml" (as required by dynamic-plugins initContainer)
This PR makes deployment fail with error message in log. 
For further improvement it:
- will be added to configuration doc ( https://github.com/janus-idp/operator/issues/129 )
- (optionally) to add somewhere to status

https://issues.redhat.com/browse/RHIDP-1740

## PR acceptance criteria

- [X] Tests


rm3l (approved):
But this will need to be cherry-picked later into the next patch release planned (1.2.1).
So we will cherry-pick this later into the `1.2.x` branch right after the 1.2.0 release of RHDH is out.

PR #383, opened 2024-06-10
add options for olm deploy and fix developer.md

gazarenkov:
## Description

This PR adds clarifications of how to deploy and test Operator on OLM for vanilla K8s and Openshift and fixes Makefile to make it possible.

## PR acceptance criteria

- [X] Documentation

Follow ./docs/developer.md -> Deploy with Operator Lifecycle Manager section to test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -363,25 +363,40 @@ release-build: bundle image-build bundle-build catalog-build ## Build operator, 
 .PHONY:
 release-push: image-push bundle-push catalog-push ## Push operator, bundle + catalog images
 
+# It has to be the same namespace as ./config/default/kustomization.yaml -> namespace
+OPERATOR_NAMESPACE ?= backstage-system
+OLM_NAMESPACE ?= olm
+
 .PHONY: deploy-olm
 deploy-olm: ## Deploy the operator with OLM
-	kubectl apply -f config/samples/catalog-operator-group.yaml
-	sed "s/{{VERSION}}/$(subst /,\/,$(VERSION))/g" config/samples/catalog-subscription-template.yaml | sed "s/{{DEFAULT_OLM_NAMESPACE}}/$(subst /,\/,$(DEFAULT_OLM_NAMESPACE))/g" | kubectl apply -f -
+	kubectl -n ${OPERATOR_NAMESPACE} apply -f config/samples/catalog-operator-group.yaml
+	sed "s/{{VERSION}}/$(subst /,\/,$(VERSION))/g" config/samples/catalog-subscription-template.yaml | sed "s/{{OLM_NAMESPACE}}/$(subst /,\/,$(OLM_NAMESPACE))/g" | kubectl -n ${OPERATOR_NAMESPACE} apply -f -
 
 .PHONY: undeploy-olm
 undeploy-olm: ## Un-deploy the operator with OLM
-	-kubectl delete subscriptions.operators.coreos.com backstage-operator
-	-kubectl delete operatorgroup backstage-operator-group
-	-kubectl delete clusterserviceversion backstage-operator.$(VERSION)
+	-kubectl -n ${OPERATOR_NAMESPACE} delete subscriptions.operators.coreos.com backstage-operator
+	-kubectl -n ${OPERATOR_NAMESPACE} delete operatorgroup backstage-operator-group
+	-kubectl -n ${OPERATOR_NAMESPACE} delete clusterserviceversion backstage-operator.v$(VERSION)
 
-DEFAULT_OLM_NAMESPACE ?= openshift-marketplace
 .PHONY: catalog-update
 catalog-update: ## Update catalog source in the default namespace for catalogsource
-	-kubectl delete catalogsource backstage-operator -n $(DEFAULT_OLM_NAMESPACE)
-	sed "s/{{CATALOG_IMG}}/$(subst /,\/,$(CATALOG_IMG))/g" config/samples/catalog-source-template.yaml | kubectl apply -n $(DEFAULT_OLM_NAMESPACE) -f -
+	-kubectl delete catalogsource backstage-operator -n $(OLM_NAMESPACE)
+	sed "s/{{CATALOG_IMG}}/$(subst /,\/,$(CATALOG_IMG))/g" config/samples/catalog-source-template.yaml | kubectl apply -n $(OLM_NAMESPACE) -f -
 
+# Deploy on Openshift cluster using OLM (by default installed on Openshift)
 .PHONY: deploy-openshift
-deploy-openshift: release-build release-push catalog-update ## Deploy the operator on openshift cluster
+deploy-openshift: OLM_NAMESPACE=openshift-marketplace release-build release-push catalog-update deploy-olm ## Deploy the operator on openshift cluster
+
+.PHONY: install-olm
+install-olm: operator-sdk
+	$(OPSDK) olm install
+
+.PHONY: create-operator-namespace
+create-operator-namespace:
+	-kubectl create namespace ${OPERATOR_NAMESPACE}
+
+.PHONY: deploy-k8s-olm
+deploy-k8s-olm: release-build release-push catalog-update create-operator-namespace deploy-olm ## Deploy the operator on openshift cluster
 
 # After this time, Ginkgo will emit progress reports, so we can get visibility into long-running tests.
 POLL_PROGRESS_INTERVAL := 120s
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -1,7 +1,8 @@
-# DEVELOPER GUIDE --- WIP
+# Developer Guide 
 
 ## Contributing
 // TODO(user): Add detailed information on how you would like others to contribute to this project
+
 
 ### How it works
 This project aims to follow the Kubernetes [Operator pattern](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/).
@@ -9,13 +10,12 @@ This project aims to follow the Kubernetes [Operator pattern](https://kubernetes
 It uses [Controllers](https://kubernetes.io/docs/concepts/architecture/controller/)
 which provides a reconcile function responsible for synchronizing resources until the desired state is reached on the cluster.
 
-
 ## Local development
 
 ### Prerequisites
 
 * **kubectl**. See [Instaling kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl).
-* **minikube**. See [Instaling minkube](https://kubernetes.io/docs/tasks/tools/#minikube).
+* Available local or remote Kubernetes cluster with cluster admin privileges. For instance **minikube**. See [Instaling minkube](https://kubernetes.io/docs/tasks/tools/#minikube).
 * A copy of the Backstage Operator sources:
 ```sh
 git clone https://github.com/janus-idp/operator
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -363,25 +363,40 @@ release-build: bundle image-build bundle-build catalog-build ## Build operator,
...
+.PHONY: install-olm
+install-olm: operator-sdk
+	$(OPSDK) olm install
+
+.PHONY: create-operator-namespace
+create-operator-namespace:
+	-kubectl create namespace ${OPERATOR_NAMESPACE}
+
+.PHONY: deploy-k8s-olm
+deploy-k8s-olm: release-build release-push catalog-update create-operator-namespace deploy-olm ## Deploy the operator on openshift cluster
```
rm3l, line 399: Why do we have this additional `deploy-k8s-olm`, while the existing `deploy-olm` seems to be sufficient?
Also, I am not sure we need to depend on `release-build` and `release-push`, as this would require the ability to push to a registry. But what if I just want to deploy using the default upstream images (from `quay.io/janus-idp`)?
I think `release-build` and `release-push` should run separately upon request.

Makefile
```diff
@@ -363,25 +363,40 @@ release-build: bundle image-build bundle-build catalog-build ## Build operator,
...
 catalog-update: ## Update catalog source in the default namespace for catalogsource
-	-kubectl delete catalogsource backstage-operator -n $(DEFAULT_OLM_NAMESPACE)
-	sed "s/{{CATALOG_IMG}}/$(subst /,\/,$(CATALOG_IMG))/g" config/samples/catalog-source-template.yaml | kubectl apply -n $(DEFAULT_OLM_NAMESPACE) -f -
+	-kubectl delete catalogsource backstage-operator -n $(OLM_NAMESPACE)
+	sed "s/{{CATALOG_IMG}}/$(subst /,\/,$(CATALOG_IMG))/g" config/samples/catalog-source-template.yaml | kubectl apply -n $(OLM_NAMESPACE) -f -
 
+# Deploy on Openshift cluster using OLM (by default installed on Openshift)
 .PHONY: deploy-openshift
-deploy-openshift: release-build release-push catalog-update ## Deploy the operator on openshift cluster
+deploy-openshift: OLM_NAMESPACE=openshift-marketplace release-build release-push catalog-update deploy-olm ## Deploy the operator on openshift cluster
```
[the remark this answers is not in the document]
rm3l: The difference between `deploy-openshift` and `deploy-olm` is only the different default OLM namespace.

docs/developer.md
```diff
@@ -1,115 +1,151 @@
...
 ```sh
-make catalog-update
+[OLM_NAMESPACE=<olm-namespace>] [IMAGE_TAG_BASE=<your-registry>/backstage-operator] make catalog-update
 ```
-4. To deloy the operator with OLM
+You can point the namespace where OLM installed. By default, in a vanilla Kubernetes, OLM os deployed on 'olm' namespace. In Openshift you have to explicitly point it to **openshift-marketplace** namespace.
+
+#### Deploy the Operator with OLM 
+Default namespace to deploy the Operator is called **backstage-system** , this name fits one defined in [kustomization.yaml](../config/default/kustomization.yaml). So, if you consider changing it you have to change it in this file and define **OPERATOR_NAMESPACE** environment variable.
+Following command creates OperatorGroup and Subscrtiption on Operator namespace
```
rm3l, line 125:
```suggestion
Following command creates OperatorGroup and Subscription on Operator namespace
```

docs/developer.md
```diff
@@ -1,115 +1,151 @@
...
 
-### Test on the local cluster
+It only takes a few seconds to run, but covers quite a lot of functionality. For early regression detection, it is recommended to run it as often as possible during development.
 
-You’ll need a Kubernetes cluster to run against.
-You can use [minikube](https://kubernetes.io/docs/tasks/tools/#minikube) or [kind](https://kubernetes.io/docs/tasks/tools/#kind) to get a local cluster for testing, or run against a remote cluster.
+### Test on the cluster
 
-**Note:** Your controller will automatically use the current context in your kubeconfig file (i.e. whatever cluster `kubectl cluster-info` shows).
+For testing, you will need a Kubernetes cluster, either remote (with sufficient admin rights) or local, such as [minikube](https://kubernetes.io/docs/tasks/tools/#minikube) or [kind](https:// kubernetes.io/docs/tasks/tools/#kind)
```
rm3l, line 38:
```suggestion
For testing, you will need a Kubernetes cluster, either remote (with sufficient admin rights) or local, such as [minikube](https://kubernetes.io/docs/tasks/tools/#minikube) or [kind](https://kubernetes.io/docs/tasks/tools/#kind)
```

docs/developer.md
```diff
@@ -1,115 +1,151 @@
...
+- Install the Custom Resource Definitions into the local cluster (minikube is installed and running):
 ```sh
 make install
 ```
+**IMPORTANT:** If you are editing the CRDs, make sure you reinstall it before deploying.
 
--  You can run your controller standalone (this will run in the foreground, so switch to a new terminal if you want to leave it running)
-This way you can see controllers log just in your terminal window which is quite convenient for debugging:
+- To delete the CRDs from the cluster:
+-
```
rm3l, line 52:
```suggestion
```

Makefile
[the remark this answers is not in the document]
rm3l: Another point: similarly to the current `undeploy-olm` target, I guess an `undeploy-openshift` target might be helpful as well, with the only difference being in the OLM namespace.

Makefile
```diff
@@ -363,25 +363,40 @@ release-build: bundle image-build bundle-build catalog-build ## Build operator,
...
+	sed "s/{{VERSION}}/$(subst /,\/,$(VERSION))/g" config/samples/catalog-subscription-template.yaml | sed "s/{{OLM_NAMESPACE}}/$(subst /,\/,$(OLM_NAMESPACE))/g" | kubectl -n ${OPERATOR_NAMESPACE} apply -f -
 
 .PHONY: undeploy-olm
 undeploy-olm: ## Un-deploy the operator with OLM
-	-kubectl delete subscriptions.operators.coreos.com backstage-operator
-	-kubectl delete operatorgroup backstage-operator-group
-	-kubectl delete clusterserviceversion backstage-operator.$(VERSION)
+	-kubectl -n ${OPERATOR_NAMESPACE} delete subscriptions.operators.coreos.com backstage-operator
+	-kubectl -n ${OPERATOR_NAMESPACE} delete operatorgroup backstage-operator-group
+	-kubectl -n ${OPERATOR_NAMESPACE} delete clusterserviceversion backstage-operator.v$(VERSION)
```
rm3l, line 379:
```suggestion
	-kubectl -n ${OPERATOR_NAMESPACE} delete clusterserviceversion backstage-operator.v$(VERSION)
	-kubectl delete catalogsource backstage-operator -n $(OLM_NAMESPACE)
```

I think we should also delete the catalog source here, to make sure there is nothing all OLM-related resources that were created have been deleted.

rm3l (changes requested):

Makefile
```diff
@@ -363,25 +363,40 @@ release-build: bundle image-build bundle-build catalog-build ## Build operator,
 .PHONY:
 release-push: image-push bundle-push catalog-push ## Push operator, bundle + catalog images
 
+# It has to be the same namespace as ./config/default/kustomization.yaml -> namespace
+OPERATOR_NAMESPACE ?= backstage-system
+OLM_NAMESPACE ?= olm
+
 .PHONY: deploy-olm
 deploy-olm: ## Deploy the operator with OLM
```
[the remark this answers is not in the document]
gazarenkov: yep, and I even created create-operator-namespace for this purpose :)

Makefile
gazarenkov: I consider it as a helper for developer, to not to keep in the brain all the steps needed, same as deploy-openshift. 
If one want to perform only certain steps - nothing prevents to. No?

Makefile
gazarenkov: catalog source is not created by deploy-olm, so it is hardly logical to put it here
we can do something like delete-catalog as complimentary action to update-catalog but I am not sure in it usefulness (+0 actually)

Makefile
```diff
@@ -363,25 +363,40 @@ release-build: bundle image-build bundle-build catalog-build ## Build operator,
...
+	-kubectl delete catalogsource backstage-operator -n $(OLM_NAMESPACE)
+	sed "s/{{CATALOG_IMG}}/$(subst /,\/,$(CATALOG_IMG))/g" config/samples/catalog-source-template.yaml | kubectl apply -n $(OLM_NAMESPACE) -f -
 
+# Deploy on Openshift cluster using OLM (by default installed on Openshift)
 .PHONY: deploy-openshift
-deploy-openshift: release-build release-push catalog-update ## Deploy the operator on openshift cluster
+deploy-openshift: OLM_NAMESPACE=openshift-marketplace release-build release-push catalog-update deploy-olm ## Deploy the operator on openshift cluster
+
+.PHONY: install-olm
+install-olm: operator-sdk
```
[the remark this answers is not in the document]
gazarenkov: Well, both of them are quite trivial, I just considered 'install-olm' is like a quick start for using olm with operator, while undeploying operator is something one should to do with care in general case. 
My position is rather "no" then "yes", so if you have additional arguments, we can discuss it further :)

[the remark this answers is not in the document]
rm3l: No special argument, but my point is that, no matter how relatively "trivial" it can be, if we expose an `install-$something` target, I would expect to see the counterpart of it. Even when I'm going through a quick start/tutorial, it usually wraps up by cleaning all the resources that were created.
If you think it is trivial enough, in this case, I would suggest even not adding this `install-olm` task, to reduce the clutter in this Makefile and keep it simple.

Makefile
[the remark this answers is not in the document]
rm3l: Are you able to run `make deploy-openshift`?

Makefile
rm3l: Yeah, but I find it redundant (and a bit confusing).. Plus, the `release-push` target requires push access to a registry.
If I just cloned the repo, and just want to deploy the operator in my cluster (where OLM is already installed) and I don't have access to a registry, which command should I run? `deploy-olm` or `deploy-k8s-olm`?
To me, only `make catalog-update deploy-olm`. (And that's why I was even suggesting that `deploy-olm` depends on `catalog-update`, so that people can just run `make deploy-olm` on K8s and have it working with the operator upstream images).
IMO, `release-build` and `release-push` should not be needed to deploy with OLM.

Makefile
rm3l: Yeah, I agree a separate target to delete the catalog source is not that useful, but I was suggesting this if we agree to make `deploy-olm` depend on `catalog-update`  (see https://github.com/janus-idp/operator/pull/383#discussion_r1636103063), in which case `undeploy-olm` should also delete the catalog source.

Makefile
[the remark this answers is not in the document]
gazarenkov: fixed

undeploy-olm will work on openshift as well

>The difference between deploy-openshift and deploy-olm is only the different default OLM namespace.

I think you are talking about deploy-openshift and deploy-k8s-olm. Yes it is the only difference. Those are helper functions. In principle deploy-k8s-olm will work for Openshift as well if set correct OLM_NAMESPACE

Makefile
gazarenkov: I believe, you, as developer, should have access to "some" registry to be able to test the full cycle, no?

As it stated in the doc "To build and deploy the operator... with OLM", so if you need build and deploy - use it, if you need just deploy - use  `catalog-update` and|or `deploy-olm` or whatever.
I find `deploy-k8s-olm` useful because I do not do it that often and I do not want to remember all the sequence every time.

Regarding whether we need separated (or dependant) `deploy-olm` and `catalog-update`, I have no strong opinion. In this case I'd just use pre-created options and see what to improve in the future.

Makefile
gazarenkov: Like I said, I have no strong opinion on making deploy-olm depend on catalog-update.
If you can tell what is the problem we are solving this way,  we can discuss (better separately from this issue), otherwise - I would use given (Note: we have those two before my PR) and see.

Makefile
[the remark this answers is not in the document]
gazarenkov: Not sure I understand your suggestion, as you can see `install-olm` is a part of "how to deploy on OLM" doc, which is the purpose of this task. 
IMO no matter how many targets we have on Makefile if they have purpose (even if redundant), are supported and documented.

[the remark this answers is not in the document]
rm3l: Okay, let's leave it as it is. We can always add it later if need be.

Makefile
rm3l: > I believe, you, as developer, should have access to "some" registry to be able to test the full cycle, no?

That makes sense if I am iterating locally and want to test everything, I agree. But I was more referring to someone who just cloned the repo and just wants to deploy using the upstream images..
Similarly to how `make deploy` just works, `make deploy-olm` should do the same thing. And if I want to build and deploy from my own images, I can run `make release-build release-push deploy-olm IMG=[my-registry/my-repo]`, but separately.
This is why I found this new `deploy-k8s-olm` to be kinda confusing.


> As it stated in the doc "To build and deploy the operator... with OLM", so if you need build and deploy - use it, if you need just deploy - use catalog-update and|or deploy-olm or whatever.
I find deploy-k8s-olm useful because I do not do it that often and I do not want to remember all the sequence every time.

In this case, maybe you could call this target differently (with `build` somewhere in the name). `deploy-k8s-olm` looks really confusing compared to `deploy-olm` ;)

> Regarding whether we need separated (or dependant) deploy-olm and catalog-update, I have no strong opinion. In this case I'd just use pre-created options and see what to improve in the future.

The current `deploy-olm` target only deploys the OperatorGroup and Subscription, which won't be useful until the catalog source is installed with `make catalog-update`. That's why I thought it would really help to add this dependency. But okay, I'm fine with what we have now.

To summarize, I think:

- `deploy-olm` is good as it is if it also depends on `catalog-update` to install the catalog source. So I am not sure we need an additional `deploy-k8s-olm` target.
- we can have a `deploy-openshift` target as helper, and the only difference would be that it would set the OLM namespace and do exactly the same thing as `deploy-olm`. No need at all to depend on the `release-*` targets, IMO.
- I am not sure we need an additional `deploy-olm-openshift`, which I also find confusing. If we have the `deploy-openshift` target above, it implies that we'll be using the OpenShift way for deploying the operator. Otherwise, there is already the `deploy` target handy.
- Not sure either we need an additional `catalog-update-openshift` target. If `deploy-openshift` just sets the expected OLM namespace and does the same thing as `deploy-olm`

This way, users can run:
- `make deploy-olm` or `make deploy-openshift` to deploy using OLM
- they can call `make release-build release-push` on demand if they need to build a different image. Then call `make deploy-olm`, `make deploy-openshift` or `make deploy` depending on what they want to test

Does that make sense?

gazarenkov: I added and documented what I found useful for developer, mostly to not to try to remember every time all the specifics.
I did not consider redundancy as a problem, to me makefile is a set of helpers.
For the deploy-olm dependency on catalog-update , like I said, I changed nothing, it is not the point of this PR. 

I believe all the "older" things should work as before.
If you see any breaking changes or documentation inconsistency, please let me know and I'll fix it ASAP, otherwise I would propose to merge it and, of course, you can modify and propose yours, more elegant, version of makefile and documentation as a separate PR.
Again, my point is to make it possible to take a glance on doc and make what I, as developer, need, e2e. 

Does it make sense?

PR #384, opened 2024-06-10
fix: Fix reconciliation error preventing existing Backstage operands from being upgraded when the operator is upgraded from `1.1.x` [RHIDP-2597]

### LINKED ISSUE #382: Existing Backstage operand not upgraded (stuck on mounting a ConfigMap) after upgrading operator from 1.1.x to 1.2.x
/kind bug

## What did you do exactly?

- From the operator repo, switch to the 1.1.x branch and deploy the operator from RHDH 1.1 (setting the `IMG` arg because the default image for this branch (`quay.io/janus-idp/operator:0.1.3`) has expired and no longer exists on quay.io)

```
git switch 1.1.x
make deploy IMG=quay.io/rhdh/rhdh-rhel9-operator:1.1
```

- Create a simple CR (e.g., [`examples/bs1.yaml`](https://github.com/janus-idp/operator/blob/1.1.x/examples/bs1.yaml))
```
kubectl apply -f examples/bs1.yaml
```

- Wait a few seconds until all the resources are created

-  Check the `Backstage` Custom Resource status. Reason should be `DeployOK`:

```
$ kubectl describe backstage bs1                                                                                                                                                           
Name:         bs1                                                                                                                                                                              
Namespace:    my-ns                                                                                                                                                                            
Labels:       <none>                                                                                                                                                                           
Annotations:  <none>                                                                                                                                                                           
API Version:  rhdh.redhat.com/v1alpha1                                                                                                                                                         
Kind:         Backstage                                                                                                                                                                        
Metadata:                                      
  Creation Timestamp:  2024-06-07T12:21:58Z                                                                                                                                                    
  Generation:          1                                                                       
  Resource Version:    48634                   
  UID:                 e4e9766f-6c32-4b44-85cd-1eac93b56f16                                                                                                                                    
Status:                                                                                                                                                                                        
  Conditions:                      
    Last Transition Time:  2024-06-07T12:21:58Z 
    Message:                                                                                                                                                                                       Reason:                DeployOK                                                                                                                                                            
    Status:                True                                                                                                                                                                
    Type:                  Deployed            
Events:                    <none>
```

- Switch to the 1.2.x branch and deploy the upcoming 1.2 operator
```
git switch 1.2.x
make deploy
```

- Wait a few seconds until the new version of the operator pod is running and the existing CR is reconciled again. Then check the CR status again. Reason will be `DeployFailed` with an error :

```
$ kubectl describe backstage bs1
Name:         bs1
Namespace:    my-ns
Labels:       <none>
Annotations:  <none>
API Version:  rhdh.redhat.com/v1alpha1
Kind:         Backstage
Metadata:
  Creation Timestamp:  2024-06-07T13:44:15Z
  Generation:          1
  Resource Version:    2846
  UID:                 5cdad2eb-0840-4c8e-8c2b-08be46e3856a
Status:
  Conditions:
    Last Transition Time:  2024-06-07T13:49:06Z
    Message:               failed to apply backstage objects failed to patch object &Service{ObjectMeta:{backstage-psql-bs1  my-ns   2100 0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[app.kubernetes.io/instance:bs1 app.kubernetes.io/name:backstage rhdh.redhat.com/app:backstage-psql-bs1] map[] [{rhdh.redhat.com/v1alpha1 Backstage bs1 5cdad2eb-0840-4c8e-8c2b-08be46e3856a 0xc000590f55 0xc000590f54}] [] []},Spec:ServiceSpec{Ports:[]ServicePort{ServicePort{Name:,Protocol:,Port:5432,TargetPort:{0 0 },NodePort:0,AppProtocol:nil,},},Selector:map[string]string{rhdh.redhat.com/app: backstage-psql-bs1,},ClusterIP:None,Type:,ExternalIPs:[],SessionAffinity:,LoadBalancerIP:,LoadBalancerSourceRanges:[],ExternalName:,ExternalTrafficPolicy:,HealthCheckNodePort:0,PublishNotReadyAddresses:false,SessionAffinityConfig:nil,IPFamilyPolicy:nil,ClusterIPs:[],IPFamilies:[],AllocateLoadBalancerNodePorts:nil,LoadBalancerClass:nil,InternalTrafficPolicy:nil,},Status:ServiceStatus{LoadBalancer:LoadBalancerStatus{Ingress:[]LoadBalancerIngress{},},Conditions:[]Condition{},},}: failed to patch object *v1.Service: Service "backstage-psql-bs1" is invalid: spec.clusterIPs[0]: Invalid value: []string{"None"}: may not change once set
    Reason:                DeployFailed
    Status:                False
    Type:                  Deployed
Events:                    <none>
```

## Actual behavior

It seems the existing CR could not be reconciled successfully with the new version of the operator because it was unable to patch the existing database Service object.

If we take a look at the resources, a new Backstage pod was in the process of being created, but is stuck on trying a mount a ConfigMap (which could not be created because of the failure to patch the DB Service):

```
$ kubectl get pod

NAME                             READY   STATUS     RESTARTS   AGE
backstage-psql-bs1-0             1/1     Running    0          9m22s
backstage-bs1-655f659ddc-n7grw   1/1     Running    0          9m22s
backstage-bs1-6469fdd48f-ldq5h   0/1     Init:0/1   0          4m31s

$ kubectl describe pod backstage-bs1-6469fdd48f-ldq5h

[...]
Events:
  Type     Reason            Age                   From               Message
  ----     ------            ----                  ----               -------
  Warning  FailedScheduling  5m18s                 default-scheduler  0/1 nodes are available: waiting for ephemeral volume controller to create the persistentvolumeclaim "backstage-bs1-6469fdd48f-ldq5h-dynamic-plugins-root". preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod..
  Normal   Scheduled         5m11s                 default-scheduler  Successfully assigned my-ns/backstage-bs1-6469fdd48f-ldq5h to k3d-k3s-default-server-0
  Warning  FailedMount       3m9s                  kubelet            Unable to attach or mount volumes: unmounted volumes=[backstage-appconfig-bs1], unattached volumes=[backstage-appconfig-bs1 dynamic-plugins-root dynamic-plugins-npmrc backstage-dynamic-plugins-bs1]: timed out waiting for the condition
  Warning  FailedMount       62s (x10 over 5m12s)  kubelet            MountVolume.SetUp failed for volume "backstage-appconfig-bs1" : configmap "backstage-appconfig-bs1" not found
  Warning  FailedMount       51s                   kubelet            Unable to attach or mount volumes: unmounted volumes=[backstage-appconfig-bs1], unattached volumes=[dynamic-plugins-root dynamic-plugins-npmrc backstage-dynamic-plugins-bs1 backstage-appconfig-bs1]: timed out waiting for the condition
```

## Expected behavior

CR reconciliation should be successful, and the application should be upgraded.

Note that the same issue happens when upgrading from the operator channels on OpenShift (https://github.com/janus-idp/operator/blob/main/.rhdh/docs/installing-ci-builds.adoc).

rm3l:
## Description

This PR fixes a reconciliation error occurring against existing CRs after upgrading the operator from the released `1.1.x` branch to either `main` or `1.2.x`.

This issue is currently due to a PATCH operation that was denied by Kubernetes against the existing database Service object.
https://github.com/janus-idp/operator/pull/369 (fix for https://issues.redhat.com/browse/RHIDP-2432) introduced a mechanism to "replace" `StatefulSet` objects when the PATCH operation fails, by deleting it (without deleting its dependents), so that the next reconciliation will recreate it.
To prevent similar issues (patching fields that are not allowed by K8s) from occurring again in the future, this PR generalizes this logic ~based on the patch error status and cause (either Field Update Forbidden or Field Value Invalid)~, i.e., it tries to delete the object anyway if patching it failed, so that it can be recreated upon the next reconciliation.

- Fixes #382 
- Fixes https://issues.redhat.com/browse/RHIDP-2597

## PR acceptance criteria

- [x] Tests
- [ ] Documentation

Repro steps provided in #382.

- To reproduce the issue using the E2E tests just run `make test-e2e-upgrade`. It will leverage the default operator image (not patched) and will not pass.
- Run `make test-e2e-upgrade IMG=quay.io/janus-idp/operator:0.3.0-pr-384` to see the E2E tests passing using the operator image from this PR.


tests/e2e/e2e_test.go
```diff
@@ -184,6 +184,72 @@ var _ = Describe("Backstage Operator E2E", func() {
...
 	})
+
+	Context("Operator upgrade with existing instances", Label("upgrade"), func() {
+		const managerPodLabel = "control-plane=controller-manager"
+
+		// 0.1.3 is the version of the operator in the 1.1.x branch
+		var fromDeploymentManifest = filepath.Join(projectDir, "tests", "e2e", "testdata", "backstage-operator-0.1.3.yaml")
+		var (
+			crName = "bs1"
+			crPath = filepath.Join(projectDir, "examples", "bs1.yaml")
```
gazarenkov, line 195: Can we make it rely on some internal script and not one from "examples/bs1.yaml" for predictability?

tests/e2e/e2e_test.go
```diff
@@ -184,6 +184,72 @@ var _ = Describe("Backstage Operator E2E", func() {
 			})
 		}
 	})
+
+	Context("Operator upgrade with existing instances", Label("upgrade"), func() {
```
gazarenkov, line 188: If the goal is to validate default config when upgrade to current version from old ones (which is a valid goal) can we consider making some tiny "framework" like (for example) iterate testing the older scripts with predictable names from a directory like
testdata/
   upgrade/
       backstage-operator-0.1.x.yaml
       backstage-operator-0.2.x.yaml
       
And do the test for each script in this directory (I would also move this upgrade test to dedicated  file and make an instruction)

This way we can easily manage the scenario by adding/removing (when the version is no longer supported) these yaml scripts.

WDYT?

Makefile
```diff
@@ -406,7 +406,11 @@ $(GINKGO): $(LOCALBIN)
 
 .PHONY: test-e2e
 test-e2e: ginkgo ## Run end-to-end tests. See the 'tests/e2e/README.md' file for more details.
-	$(GINKGO) $(GINKGO_FLAGS) tests/e2e
+	$(GINKGO) $(GINKGO_FLAGS) --label-filter="!upgrade" tests/e2e
```
gazarenkov, line 409: Curious: What is the reason to separate it this way?

gazarenkov (changes requested):

Makefile
rm3l: The "normal" tests have a `BeforeSuite` logic that deploys the operator image once in the cluster (from an end-user perspective) before running the tests.
The "upgrade" test is different because it needs a different version (older) version of the operator to be installed. As such, it cannot share the same controller as the other ones.

tests/e2e/e2e_test.go
rm3l: Yeah, that makes sense. I thought about a similar approach when adding this `backstage-operator-x.y.z.yaml` file and including such "upgrade" tests as part of the nightly Workflow. But due to the current severity of this bug, I think this can be a later improvement, which can even be assigned to one of our newcomers to get started.
I'll create a separate issue for this.

controllers/backstage_controller.go
```diff
@@ -181,11 +181,12 @@ func (r *BackstageReconciler) applyObjects(ctx context.Context, objects []model.
 		}
 
 		if err := r.patchObject(ctx, baseObject, obj); err != nil {
-			_, isStatefulSet := obj.Object().(*appsv1.StatefulSet)
-			if isStatefulSet && errors.IsInvalid(err) && errors.HasStatusCause(err, metav1.CauseTypeForbidden) {
-				// Some resources like StatefulSets allow patching a limited set of fields.
-				// That's why we are trying to delete them first, taking care of orphaning the dependents so that resources like PVCs can be retained.
+			if errors.IsInvalid(err) && (errors.HasStatusCause(err, metav1.CauseTypeForbidden) || errors.HasStatusCause(err, metav1.CauseTypeFieldValueInvalid)) {
```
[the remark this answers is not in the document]
rm3l: Yeah, I think it won't hurt to unconditionally try to recreate the object anyway in this case.

tests/e2e/e2e_test.go
rm3l: What do you mean by "some internal script"? To rewrite the whole test as a script?
We are supposed to maintain the example CRs, so it made sense to me for our E2E tests to leverage them.

controllers/backstage_controller.go
```diff
@@ -181,11 +181,12 @@ func (r *BackstageReconciler) applyObjects(ctx context.Context, objects []model.
...
+				// Some resources like StatefulSets allow patching a limited set of fields. A FieldValueForbidden error is returned.
+				// Some other resources like Services do not support updating the primary/secondary clusterIP || ipFamily. A FieldValueInvalid error is returned.
+				// That's why we are trying to delete them first, taking care of orphaning the dependents so that they can be retained.
 				// They will be recreated at the next reconciliation.
+				// If they cannot be recreated at the next reconciliation, the expected error will be returned.
 				if err = r.Delete(ctx, baseObject, client.PropagationPolicy(metav1.DeletePropagationOrphan)); err != nil {
 					return fmt.Errorf("failed to delete object %s so it can be recreated: %w", obj.Object(), err)
 				}
```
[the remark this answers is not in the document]
rm3l: Done in [`e9ff462` (#384)](https://github.com/janus-idp/operator/pull/384/commits/e9ff462b926142911c9022b46af3e806a7e1f135)

controllers/backstage_controller.go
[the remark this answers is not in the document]
rm3l: Done in [`e9ff462` (#384)](https://github.com/janus-idp/operator/pull/384/commits/e9ff462b926142911c9022b46af3e806a7e1f135)

tests/e2e/e2e_test.go
gazarenkov: Nope, I meant yaml script, to use dedicated one, not from examples.

Makefile
gazarenkov: OK, then probably it makes sense to implement it in separate file?
To me it makes sense to separate it b/c I suppose there are no reason to call it unless you change default configuration (more rare case than regression caused by code change), so it can be release related at all.

controllers/backstage_controller.go
```diff
@@ -181,22 +181,22 @@ func (r *BackstageReconciler) applyObjects(ctx context.Context, objects []model.
...
+			lg.V(1).Info("failed to patch object => trying to delete it so it can be recreated upon next reconciliation...",
+				objDispName(obj), obj.Object().GetName(),
+				"cause", err,
+			)
+			// Some resources like StatefulSets allow patching a limited set of fields. A FieldValueForbidden error is returned.
+			// Some other resources like Services do not support updating the primary/secondary clusterIP || ipFamily. A FieldValueInvalid error is returned.
+			// That's why we are trying to delete them first, taking care of orphaning the dependents so that they can be retained.
+			// They will be recreated at the next reconciliation.
+			// If they cannot be recreated at the next reconciliation, the expected error will be returned.
+			if err = r.Delete(ctx, baseObject, client.PropagationPolicy(metav1.DeletePropagationOrphan)); err != nil {
```
gazarenkov, line 193: Just realized, seems we loose one functionality here.
patchObject() also saves labels and annotations which would be lost on delete I think.
Thoughts?

gazarenkov (changes requested):

tests/e2e/e2e_test.go
rm3l: Okay, I can do that, but still don't understand the benefit. In the scope of end-to-end tests, I want to make sure the examples are maintained and work, and having one of them used for testing the upgrade path here seems to make sense to me, doesn't it?

controllers/backstage_controller.go
rm3l: Do you mean if someone manually adds labels and annotations to the resources created by the operator? But is that case something we want to recommend?
I guess they should add such labels and annotations via the raw config instead..

rm3l: But it is probably safer in this case to revert [`e9ff462` (#384)](https://github.com/janus-idp/operator/pull/384/commits/e9ff462b926142911c9022b46af3e806a7e1f135) then, and delete the object only on known error causes. That doesn't prevent losing additional labels/annotations, but at least it limits the scope to specific errors.

tests/e2e/e2e_test.go
gazarenkov: Well, to me tests and example it is 2 different things :) 
Your point makes sense too...
OK, it is a different topic, let's go as it is.

Makefile
rm3l: Well, the original issue was caused by a regression due to some code change (change to the DB Service object spec).
But we should capture those upgrade issues early on, be it release or code-related, and that's why I think it will be helpful to have them run regularly.
I am using a label decorator here as a way to separate those, but okay, I can move those to a dedicated spec file.

controllers/backstage_controller.go
rm3l: Okay, let's call this out in the release notes.

Makefile
rm3l: Moved to a separate spec file. See [`077da8a` (#384)](https://github.com/janus-idp/operator/pull/384/commits/077da8a60c37adde9a978562b7e7559932cbce02)

controllers/backstage_controller.go
gazarenkov: And please add some Note that annotations and labels will be lost somewhere here https://github.com/janus-idp/operator/blob/077da8a60c37adde9a978562b7e7559932cbce02/controllers/backstage_controller.go#L196 or to the message above.

rm3l: Added in [`6d2fa79` (#384)](https://github.com/janus-idp/operator/pull/384/commits/6d2fa79077bb92c6cf6bb6a8370bb7565d696bca)

gazarenkov (approved):

PR #392, opened 2024-07-01
[1.2.x] Allow configuring the storage class name for PVCs used by both DB and app

rm3l:
## Description

This PR adds 2 new fields to the CRD:
- `spec.application.storageClassName`
- `spec.database.storageClassName`

They allow configuring the storage classes to be used by the PVCs created by the Operator, i.e., the ephemeral PVC created for the `dynamic-plugins-root` volume, as well as the PVC used by the local database StatefulSet.

https://github.com/janus-idp/operator/pull/388 intended to provide a more generic way to handle this (along with supporting customizing other properties like labels, annotations, resources, sidecar containers, ...), but it currently does not apply to the local database.

Given that https://issues.redhat.com/browse/RHIDP-2705 is a current a blocker for the upcoming 1.2.2 release, this PR is more focused, covering only the storage class use case with the current `v1alpha1` API version.

https://github.com/janus-idp/operator/pull/388 will need to be updated later for 1.3.

**Preview in the OCP console**
Screenshot from 2024-07-01 12-56-19

- Fixes https://issues.redhat.com/browse/RHIDP-2705

## PR acceptance criteria

- [x] Tests
- [ ] Documentation

(Tested on a ClusterBot cluster, but feel free to use any storage class on your particular cluster)
```yaml
cat <<EOF | kubectl apply -f -
apiVersion: rhdh.redhat.com/v1alpha1
kind: Backstage
metadata:
  name: developer-hub
spec:
  application:
    storageClassName: "gp2-csi"
  database:
    storageClassName: "gp3-csi"
EOF
```

Then run `kubectl get pvc` and check the storage classes used by the PVCs, e.g.:

```
$ kubectl get pvc
NAME                                                               STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
backstage-backstage-sample-5dbd785b5c-6h7qt-dynamic-plugins-root   Bound    pvc-06f259dc-6c02-4fbd-a6d1-4a78d40ca9ac   2Gi        RWO            gp2-csi        4m43s
data-backstage-psql-backstage-sample-0                             Bound    pvc-2e44619b-cf1c-4d62-bc4c-fea423537648   1Gi        RWO            gp3-csi        4m43s
```

### DIFF
--- a/.rhdh/bundle/manifests/rhdh-operator.csv.yaml
+++ b/.rhdh/bundle/manifests/rhdh-operator.csv.yaml
@@ -56,6 +56,21 @@ spec:
       displayName: Red Hat Developer Hub
       kind: Backstage
       name: backstages.rhdh.redhat.com
+      specDescriptors:
+      - description: Name of the storage class to use for the Persistent Volumes requested
+          by the application. The default storage class name will be used if this
+          field is not specified.
+        displayName: Storage Class for Application volumes
+        path: application.storageClassName
+        x-descriptors:
+        - urn:alm:descriptor:io.kubernetes:StorageClass
+      - description: Name of the storage class to use for the database Persistent
+          Volumes. The default storage class name will be used if this field is not
+          specified.
+        displayName: Storage Class for database volumes
+        path: database.storageClassName
+        x-descriptors:
+        - urn:alm:descriptor:io.kubernetes:StorageClass
       version: v1alpha1
   description: |
     Red Hat Developer Hub is an enterprise-grade platform for building developer portals, containing a supported and opinionated framework. By implementing a unified and open platform designed to maximize developer skills, ease onboarding, and increase development productivity, focus can be centered on what really matters: writing great code. Red Hat Developer Hub also offers Software Templates to simplify the development process, which can reduce friction and frustration for development teams, boosting their productivity and increasing an organization's competitive advantage.
--- a/api/v1alpha1/backstage_types.go
+++ b/api/v1alpha1/backstage_types.go
@@ -70,6 +70,7 @@ type Database struct {
 	// Name of the storage class to use for the database Persistent Volumes.
 	// The default storage class name will be used if this field is not specified.
 	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Storage Class for database volumes",xDescriptors={"urn:alm:descriptor:io.kubernetes:StorageClass"}
 	StorageClassName *string `json:"storageClassName,omitempty"`
 }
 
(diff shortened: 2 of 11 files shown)


gazarenkov (changes requested):
I do not think new feature, especially related to API change can be a part of patch version (see https://semver.org/).

Instead, to make deployment configuration feature to be available sooner, I would propose to make a 0.3 (1.3) operator release, including https://github.com/janus-idp/operator/pull/388,  ASAP.

rm3l:
> I do not think new feature, especially related to API change can be a part of patch version (see [semver.org](https://semver.org/)).
> 
> Instead, to make deployment configuration feature to be available sooner, I would propose to make a 0.3 (1.3) operator release, including #388, ASAP.

Sure, it can be seen as a feature semantically-speaking, but in my understanding, https://issues.redhat.com/browse/RHIDP-2705 has been marked as a blocker for 1.2.2, upon customer request. So it cannot wait for 1.3, I guess.

#388 covers too many things right now; so to minimize the risks of introducing breaking changes to 1.2.2, it was concluded that a dedicated PR (addressing only the storage class request) would be created for 1.2.2, hence this PR.
This PR only targets `1.2.x`, so only for the upcoming `1.2.2` release, which will leave time to address the review comments on #388 for 1.3
Hope that clarifies.

gazarenkov:
> > I do not think new feature, especially related to API change can be a part of patch version (see [semver.org](https://semver.org/)).
> > Instead, to make deployment configuration feature to be available sooner, I would propose to make a 0.3 (1.3) operator release, including #388, ASAP.
> 
> Sure, it can be seen as a feature semantically-speaking, but in my understanding, https://issues.redhat.com/browse/RHIDP-2705 has been marked as a blocker for 1.2.2, upon customer request. So it cannot wait for 1.3, I guess.
> 
> #388 covers too many things right now; so to minimize the risks of introducing breaking changes to 1.2.2, it was concluded that a dedicated PR (addressing only the storage class request) would be created for 1.2.2, hence this PR. This PR only targets `1.2.x`, so only for the upcoming `1.2.2` release, which will leave time to address the review comments on #388 for 1.3 Hope that clarifies.

No, it does not, sorry :) 

- #388 was urgently introduced as a reaction on exactly the same challenge, it is here for about 3 weeks already. It does NOT cover many things. Majority of files with code changes are related to renaming the alias (as I explained earlier) which one time operation (will not be needed until we come to Major version 2, which is highly unlikely) and is NOT risky at all.
- Perfect quality is not guaranteed in any case but this PR contains unit and integration tests for the new functionality and I made a couple of commits as a reaction on review (thanks for that) BUT we can not be blocked with new things just because we are afraid of bugs.
- We can not introduce new functionality and especially API changes in the bug-fix version. If we are blocked for making 1.3 release earlier than September than we have a real issue IMO, let's discuss and try to solve it ASAP.

rm3l:
Customer confirmed this is not needed in 1.2.x

/close

PR #390, opened 2024-06-28
Remove deprecation warning about backend.auth.keys

04kash:
## Description

Fixed default configuration to remove deprecation warning about backend.auth.keys by replacing it with https://github.com/backstage/backstage/blob/master/beps/0007-auth-external-services/README.md#the-legacy-access-type 
## Which issue(s) does this PR fix or relate to

Fixes: https://issues.redhat.com/browse/RHIDP-2716

## PR acceptance criteria

- [ ] Tests
- [ ] Documentation
- [ ] If the bundle manifests have been updated, make sure to review the [`rhdh-operator.csv.yaml`](../.rhdh/bundle/manifests/rhdh-operator.csv.yaml) file accordingly

PR #397, opened 2024-07-05
ci: Run E2E tests against all versions of K8s we claim support for

rm3l:
## Description

To help us detect potential issues in advance, this expands the E2E configuration matrix to test several versions of K8s:
- the versions of K8s that map to the OCP versions (https://access.redhat.com/solutions/4870701) that we explicitly claim support for - see https://access.redhat.com/support/policy/updates/developerhub
- some released versions of K8s that are not supported yet. in this case, they are marked as experimental to not fail the Workflow

&mdash;

## PR acceptance criteria

- [x] Tests
- [ ] Documentation

Example of run here: https://github.com/rm3l/janus-idp-operator/actions/runs/9809691738

The failures are not related to the changes here, but to an issue with the default [`janus-idp/backstage-showcase`](https://quay.io/repository/janus-idp/backstage-showcase?tab=tags) image's `latest` tag being outdated, and recently-merged https://github.com/janus-idp/operator/pull/390 only works in 1.2+. See https://redhat-internal.slack.com/archives/C04CUSD4JSG/p1720191475569659

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -8,14 +8,50 @@ on:
 
 jobs:
   e2e-tests:
+    name: 'E2E Tests (${{ matrix.k8s_version }}) - ${{ matrix.branch }}'
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
       matrix:
         branch: [ main, 1.2.x, 1.1.x ]
-    name: E2E Tests - ${{ matrix.branch }}
+        experimental: [ false ]
+        k8s_version:
+          # See https://access.redhat.com/solutions/4870701 for the mapping between OpenShift 4.x and Kubernetes versions.
+          # Per https://access.redhat.com/support/policy/updates/developerhub, we want support for OCP 4.12+, so K8s v1.25+
+          - v1.25.16 # OCP 4.12
+          - v1.26.15 # OCP 4.13
+          - v1.27.15 # OCP 4.14
+          - v1.28.11 # OCP 4.15
+        include:
+          # These are upcoming versions that we don't explicitly claim support for but still want to test against.
+          # Marked as experimental so as not to fail the build, but we still want to capture any useful information in advance in case of failures.
+          - k8s_version: v1.29.6 # OCP 4.16 (not claimed as supported yet)
+            branch: main
+            experimental: true
+          - k8s_version: latest
+            branch: main
+            experimental: true
+          - k8s_version: v1.29.6 # OCP 4.16 (not claimed as supported yet)
+            branch: 1.2.x
+            experimental: true
+          - k8s_version: latest
+            branch: 1.2.x
+            experimental: true
+          - k8s_version: v1.29.6 # OCP 4.16 (not claimed as supported yet)
+            branch: 1.1.x
+            experimental: true
+          - k8s_version: latest
+            branch: 1.1.x
+            experimental: true
+        exclude:
+          # These are the combinations not supported at all and should be excluded
+          - branch: 1.2.x
+            k8s_version: v1.25.16 # OCP v4.12 not supported in 1.2
+          - branch: main
+            k8s_version: v1.25.16 # OCP 4.12 not supported in 1.3+
     concurrency:
-      group: ${{ github.workflow }}-${{ matrix.branch }}
+      group: '${{ github.workflow }}-${{ matrix.branch }}-${{ matrix.k8s_version }}'
       cancel-in-progress: true
     env:
       CONTAINER_ENGINE: podman
@@ -49,9 +85,10 @@ jobs:
 
       - name: Start Minikube
         if: ${{ steps.operator-image-existence-checker.outputs.OPERATOR_IMAGE_EXISTS == 'true' }}
-        uses: medyagh/setup-minikube@317d92317e473a10540357f1f4b2878b80ee7b95 # v0.0.16
+        uses: medyagh/setup-minikube@d8c0eb871f6f455542491d86a574477bd3894533 # v0.0.18
         with:
           addons: ingress
+          kubernetes-version: ${{ matrix.k8s_version }}
 
       - name: Run E2E tests (Operator Upgrade path)
         # Testing upgrade from 1.1.x


gazarenkov (changes requested):
There are a couple of issues to me:

- It is not clear what issue we are solving here and so the "status" of this testing is not clear as well. It refers to OCP versions which clearly based on certain K8s version, but the tests are running against K8s (sure) so it can not be taken as OCP compatibility tests anyway.
- It definitely brings some useful compatibility information for K8s X.Y version, while Z version can be basically "any" (no breaking changes expected). Taking into account that K8s releases every 3-4 months, to me it sounds like overkill to make it nightly.
- In my opinion having them nightly just overloads the actions output (and calculation resources :) ) with mostly the same information.

So, I would say it as a useful test but we probably need to have more clear definition of it. For example: as an upstream pre-release tests which makes sure we support supportable K8s versions and happened as a part of pre-release procedure and not nightly.

rm3l:
> There are a couple of issues to me:
> 
> * It is not clear what issue we are solving here and so the "status" of this testing is not clear as well. It refers to OCP versions which clearly based on certain K8s version, but the tests are running against K8s (sure) so it can not be taken as OCP compatibility tests anyway.

I would ideally run those tests directly on OCP if it was as easy as a GitHub Action, but I couldn't find a simple way to spin up a minimal OCP cluster in a GitHub runner. So this uses the OCP versions we claim support for as baselines to determine the versions of K8s we could/should test against. 

> * It definitely brings some useful compatibility information for K8s X.Y version, while Z version can be basically "any" (no breaking changes expected). Taking into account that K8s releases every 3-4 months, to me it sounds like overkill to make it nightly.

The variable here is the operator code (which can change frequently), which I think we should have real E2E tests for, but against the versions of K8s that we should support (this list won't change frequently either).
There is also one useful part here in this Workflow IMO: ahead-of-time testing against versions of K8s that we do not support yet will help us catch potential issues in advance.

> * In my opinion having them nightly just overloads the actions output (and calculation resources :) ) with mostly the same information.
> 

The actions output are separated per job, so per configuration in the matrix. So it won't change that much from the [existing nightly workflow](https://github.com/janus-idp/operator/actions/runs/9865657473). It is just we have new jobs from the matrix to look at.
In terms of resources, all matrix configuration jobs run in parallel, so there is no impact on the overall time of the whole Workflow. The [current nightly workflow](https://github.com/janus-idp/operator/actions/runs/9865657473/usage) on `main` takes ~19m to run, and the [Workflow from this PR](https://github.com/rm3l/janus-idp-operator/actions/runs/9809691738/usage) took almost the same amount of time.

> So, I would say it as a useful test but we probably need to have more clear definition of it. For example: as an upstream pre-release tests which makes sure we support supportable K8s versions and happened as a part of pre-release procedure and not nightly.

The goal of the nightly Workflow is to prevent a last-minute rush before a release when you notice something does not work end-to-end, by running a comprehensive suite of tests against the released image regularly. This way, based on your last nightly job status, you can even release any good commit at any time.

Are you concerned about having to check the status too frequently?

coreydaley:
You can setup tests on OpenShift CI to run against OpenShift clusters, there are lots of examples of how to do this in the github.com/openshift org, specifically look at the https://github.com/openshift/release repository

rm3l:
> Like I said, the compatibility tests (k8s or/and OCP if possible) are useful.
> 
> At the same time in absolute majority of the cases they do not bring any useful information daily, so very quickly will be percieved as a "white noice", as a related example you can see current state of nightly tests, they failed in many cases by known reason but we do not care about it b/c they are not that important b/c pretty much everything is covered in other (PR) flows. On the other hand, from outside, it looks like as broken environment. That's why I really doubt in usefulness of daily tests in our context(!) at all.
> 

PR tests currently do not run against a real cluster to make sure that everything work end-to-end. So to me, I am not confident enough that a merged PR with the current tests passing guarantees that the deployed application is actually reachable (which is what matters to the end-user, right?).
Take https://github.com/janus-idp/operator/pull/390 for example. All PR tests passed, but the nightly job that ran against it allowed us to detect the day after an issue with the `latest` tag of the Showcase application (because the pod deployed ended up crashing on a real cluster).
I would rather not accumulate a lot of changes and run a manual E2E test a few days before a release..

> So, I see such tests rather as a part of the (conscious) [pre/after] release process/recommendations than periodical job.
> 
Is there such a process in place? A recommendation is only useful if there is a way to enforce it, and a nightly job is an almost free and automated way to enforce that.  ;)

> I do not really understand your "last-minute rush" concern, it's time to focus on development and time to focus on release ("code freeze" state in our case). Could you please be more specific about the cases you have in mind?

If we want to release at any time, nightly tests will help for example.

gazarenkov (changes requested):
I see this compatibility testing as good to have for sure but:

The state of current nightly tests in general is not satisfable. They are not stable for a quite long time and so looks broken.
https://github.com/janus-idp/operator/issues/399
That's why extending it with any new functionality is clearly **no go** from my side.

@rm3l The point about not having "real cluster tests" on CI (except ones from nightly) is **valid** and we have to solve it ASAP. 
After that I do not see any reason to keep nightly jobs until we are sure we are able properly maintain it and see real needs in it.

rm3l:
> I see this compatibility testing as good to have for sure but:
> 
> The state of current nightly tests in general is not satisfable. They are not stable for a quite long time and so looks broken. #399 That's why extending it with any new functionality is clearly **no go** from my side.
> 

I already stated that they are not passing (only since the past few days) because of https://github.com/janus-idp/operator/pull/390 which was recently merged. The issue (thankfully caught quickly by the nightly checks) is being tracked in this JIRA: https://issues.redhat.com/browse/RHIDP-3157
Before that, they have been quite stable so far. From time to time, some infra issues due to the runner itself, but nothing critical per se.

Anyway, let's bring this discussion in one of our WG calls, to have more opinions on how to approach this.

/hold

gazarenkov:
> > I see this compatibility testing as good to have for sure but:
> > The state of current nightly tests in general is not satisfable. They are not stable for a quite long time and so looks broken. #399 That's why extending it with any new functionality is clearly **no go** from my side.
> 
> I already stated that they are not passing (only since the past few days) because of #390 which was recently merged. The issue (thankfully caught quickly by the nightly checks) is being tracked in this JIRA: https://issues.redhat.com/browse/RHIDP-3157 Before that, they have been quite stable so far. From time to time, some infra issues due to the runner itself, but nothing critical per se.

Looks like we are talking different issues.

Take a look here:

https://github.com/janus-idp/operator/actions?page=2&query=event%3Aschedule

And you can see that majority of Nightly builds are failed.
Do you mean something else saying it is not a problem?

> 
> Anyway, let's bring this discussion in one of our WG calls, to have more opinions on how to approach this.
> 
> /hold

PR #407, opened 2024-07-19
RHIDP-1148-link-external-db-note

Fortune-Ndlovu:
Instead of adding a lengthy link at the end of the description, it is more useful to add a markdown link at the end of the More Information heading's list of markdown links.

e


.rhdh/bundle/manifests/rhdh-operator.csv.yaml
```diff
@@ -82,7 +80,8 @@ spec:
     * [Product Documentation](https://access.redhat.com/documentation/en-us/red_hat_developer_hub)
     * [Life Cycle](https://access.redhat.com/node/7025299)
     * [Support Policies](https://access.redhat.com/policy/developerhub-support-policy)
-    
+    * [Configuring external PostgreSQL databases](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.2/html/administration_guide_for_red_hat_developer_hub/assembly-configuring-external-postgresql-databases#assembly-configuring-external-postgresql-databases)
```
rm3l, line 83:
```suggestion
    * [Configuring external PostgreSQL databases](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.3/html/administration_guide_for_red_hat_developer_hub/assembly-configuring-external-postgresql-databases#assembly-configuring-external-postgresql-databases)
```

Similar to the Telemetry link (Telemetry data collection above), I guess this should be [1.3 in `main`](https://github.com/janus-idp/operator/blob/main/.rhdh/bundle/manifests/rhdh-operator.csv.yaml#L77) and [`1.2` in `1.2.x`](https://github.com/janus-idp/operator/blob/1.2.x/.rhdh/bundle/manifests/rhdh-operator.csv.yaml#L69)..

Fortune-Ndlovu: Yes that is correct I have updated this, thank you.

rm3l (approved):

